PR #4, opened 2022-08-05
Rpip 8 (LEB8s)

Valdorff:


VVander:
Great to see some engagement on the RPIP repo! 

I think these edits are small enough that it's fine, but in the future, I suggest breaking them up into multiple PRs @Valdorff .

darrenlangley:
Good summary of the LEB8 discussion. We will still have to evaluate based when we go into design but this forms a good starting point.

PR #5, opened 2022-08-09
Minor tweaks

Valdorff:
- Changed status to Final
- Clarified math in Motivation section (and corrected current locked amount)
- Added minor context on rETH apr
- Fixed a couple of typos, including one breaking header table display


james-beep-boop:
In the “Increase the Deposit Pool Limit” snapshot, change:

Increasing the deposit pool limit does present a small potential reduction risk on the rETH APR but only occurs when the deposit pool has a positive balance

To:

Increasing the deposit pool limit reduces rETH APR by (ETH in dp / total rETH TVL in ETH) when the deposit pool has a positive balance

And 

In the first sentence of the linked explanation of the proposal, change the term “unstated” to “unstaked”

Valdorff:
I think both of @james-beep-boop 's points have been addressed :)

PR #24, opened 2022-09-11
Initial PGM charter draft

Valdorff:


VVander:
Hey @Valdorff even if we disagree on the implications (maybe especially because there is disagreement), I recommend putting a section on Security Considerations in here.

Valdorff:
The onchain pDAO stuff coming in Houston supersede this.

PR #27, opened 2022-10-03
Add Final state requirement for snapshot votes

VVander:
I believe this is within the scope of a small clarifying edit (it's only one word!), but proposing even small changes to an RPIP which has passed a vote is pretty significant. Hopefully, everyone thinks it's reasonable, otherwise I can write a separate RPIP containing this edit and we can discuss + vote on it as usual. 

The reason I submitted this is because, as written before this PR, RPIP-4 can be interpreted as allowing RPIPs to be voted in during any state, e.g. "Draft" where significant content changes are still possible. I can't imagine anyone voting for something like this, so I suggest we clarify that we will only do snapshot votes on Final RPIPs.

### DIFF
--- a/RPIPs/RPIP-4.md
+++ b/RPIPs/RPIP-4.md
@@ -21,7 +21,7 @@ Governance in Rocket Pool is currently ad hoc, and there is no way for the commu
 ### Poll Procedure
   
 All Rocket Pool governance proposals MUST have an associated post on the governance forums at dao.rocketpool.net.
-Promising community sentiment and a corresponding Final RPIP document are REQUIRED for snapshot voting to be scheduled.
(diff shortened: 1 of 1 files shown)


VVander:
After some discussion on discord, this may not be sufficient since there may be votes on "Living" RPIPs as well. Therefore, we need a solution for how "Living" RPIPs should be updated.

Valdorff:
@VVander How about "Promising community sentiment and a corresponding RPIP document successfully past the Review stage are REQUIRED for snapshot voting to be scheduled."

VVander:
@Valdorff I think that works. That wouldn't solve all the problems surrounding Living RPIPs, but that's probably a separate conversation.

I'm en route to devcon but will update the PR when I can. I've also set it to allow maintainers to edit in case anyone else wants to do it.

PR #45, opened 2023-04-26
Resignation of Wander

VVander:


Valdorff:
Closing this. This is a moment in time, so don't think it needs a permanent RPIP slot. Note that folks coming from <https://dao.rocketpool.net/t/rpip-editor-resignation/1715> will still see the contents fine here, even when closed.

PR #48, opened 2023-04-26
Fix YAML error on author line

xrchz:
This removes the error on GitHub's rendering of the page. It does not actually fix all the yaml syntax errors in the file (many of which are "line too long")...


Valdorff (approved):
LGTM - 2 character change in author line to prevent header table issues

PR #61, opened 2023-07-31
Updating Membership in RPIP-18.md

prose11:
Kevster steps down, anisoptera.eth joins based on last vote.

Saw @shfryn's post and figured it might be helpful to do the PR :) let me know if any other info is required @Valdorff 

Note: used today's date since that's when it the announcement up on the forum


Valdorff (approved):
Looks good - thanks

PR #62, opened 2023-08-04
Add 2023-07-17 vote info

Valdorff:
Add vote info from the recent vote cycle to the relevant RPIPs


htimsk (approved):
Looks good. Vote results from 7/17 were added.

PR #65, opened 2023-08-20
Set RPIP-27 to Review and update RPL amount

knoshua:


xrchz (approved):
LGTM, though maybe good if another editor checks this one too 😉

Valdorff:
Looks good, checked the 2 simple bits of math and the price on July 6.
Will be finalizing and merging.

PR #67, opened 2023-08-22
Updated Living GMC with August awards

shfryn:


shfryn:
LGTM

Valdorff (approved):
Links look helpful, and tables display a bit better without the long addresses. Ty.

PR #69, opened 2023-09-01
Rpl staking rework

Valdorff:
Whoops. Didn't realize I hadn't gotten this draft into main repo.


RPIPs/RPIP-30.md
```diff
@@ -0,0 +1,123 @@
...
+and Reserves combined) has been significantly reduced so that we can spend a lot more on achieving
+the protocol goal of attracting minipool creation (and rETH supplying).
+
+## Specification
+
+Definition: X=1 is the first reward snapshot after (a) the end of the vote and (b) the
+implementation of the new rewards rules. X=2 is the snapshot after that, etc.
+
+- The new rewards rules SHALL be phased in such that, for reward snapshot X:
+  - For periods X=1 to X=5: a nodes' share of rewards is
```
xrchz, line 54: should be node's rather than nodes' I think

RPIPs/RPIP-30.md
```diff
@@ -0,0 +1,123 @@
...
+    `(X/6)*proposed_method_share + ((6-x)/6)*current_method_share`
+  - For periods X>5, they SHALL  simply be `proposed_method_share`
+  - `current_method_share` SHALL determine share of NO rewards among NOs according to the latest tree
```
xrchz, line 57: so this RPIP prevents any changes to the treegen spec for 5 periods?

Valdorff: That was not the intent, but I can see how that would be read that way. I'll add an explicit "MAY" be changed following existing processes for both current and proposed methods.

xrchz (approved):

PR #77, opened 2023-09-22
Draft RPIPs for new RPL withdrawal address and stake ETH on behalf of

darrenlangley:


Valdorff:
Quick skim thoughts for withdrawal address:
- Could we go with `ETH withdrawal address` and `RPL withdrawal address` everywhere? Or perhaps `primary withdrawal address` and `RPL withdrawal address`? The spec is tough to parse with pure `withdrawal address`
- Needs some care around node operator vs node -- I like the "as a node operator" actor part, but a withdrawal address should belong to the node not the operator
- The fact that RPL flows to primary address is only there as a subclause of unsetting - I think the flows should be primary bullets

Taking a stab at spec:
```
- The withdrawal address that has existed to date SHALL be known as the primary withdrawal address
- If a node's RPL withdrawal address is set, all RPL rewards and withdrawn RPL will go to the RPL withdrawal address
- If a node's RPL withdrawal address is unset, all RPL rewards and withdrawn RPL will go to the primary withdrawal address
- As a node operator, I MUST be able to set an RPL withdrawal address so that I can direct my RPL bond/rewards
  - If a node's RPL withdrawal address is unset, the change MUST come from the node's primary withdrawal address
  - If a node's RPL withdrawal address is set, the change MUST come from the current RPL withdrawal address
- As a node operator, I MUST be able to confirm the RPL withdrawal address from the new target address before it takes effect, to ensure it is the correct address
- As the owner of the RPL withdrawal address, I MUST be able to unset my RPL withdrawal address
- As the owner of the RPL withdrawal address, I MUST be able to call `RocketNodeStaking.withdrawRPL()` to withdraw excess RPL
- The RPL withdrawal address SHALL be initialized as unset for new node operators
- The RPL withdrawal address SHALL be initialized as unset for existing node operators
```

Valdorff:
And some thoughts for Stake ETH on behalf:

- Should we consider SHOULD language about this flow being the preferred flow if the node address is a hot wallet?
- I liked your "As the owner of X address" from the other rpip better than "as X address"
- Your last bullet trying to clarify that if the withdrawal address changes then that changes who has access to unused ETH makes it muddier imo. I tried different wordings and sub-bullets and couldn't get clearer than removing it. I think it's unambiguous with "the owner of a node's withdrawal address" as this is a unique current thing.
- I think if both rpips are included, this would read better saying "primary withdrawal address". I think this would be worth calling out somewhere... maybe the related rpips?

Minor spec suggestions:
```
- As the owner of a node's withdrawal address, I MUST be able to fund the corresponding node with ETH that is accounted for separately from the node's funds
- As the owner of a node's withdrawal address, I MUST be able to withdraw unused withdrawal address supplied ETH
- As a node operator, I MUST NOT be able to withdraw the funds supplied by the withdrawal address
- As a node operator, I MUST be able to fund new minipools using the withdrawal address supplied ETH 
```

```
## Other related RPIPs 
- [**RPIP-3** - SaaS Provider Standard](./RPIP-3.md) 
- [**RPIP-7** - Withdrawal Address Escrow](./RPIP-7.md) 
- [RPIP-TBD - RPL Withdrawal Address](RPIPs/RPIP-rpl-withdrawal-address.md)
  - If the `RPL withdrawal address` RPIP is ratified, note that withdrawal address in this RPIP's Specification refers to the "primary withdrawal address". In this case, this line should be removed and the wording updated in the Specification.
```

----
One higher level thought: why can only the withdrawal address owner stake on behalf? It's mentioned as security, but I'm not really seeing the issue with staking ETH from another wallet. Certainly it gives up ownership of the ETH to the holder of the withdrawal address, but that could easily be intended (eg, I prefer not to use cold.valdorff.eth for interacting with smart contracts, but I do want it to be my withdrawal address).

darrenlangley:
> why can only the withdrawal address owner stake on behalf?

Kane and I discussed this quite a bit. We leaned towards safe-by-default. As you said, you are giving your ETH to the withdrawal address owner. At the time, we couldn't think of a strong use case that demanded not-safe-by-default.

kanewallmann:
I like "primary withdrawal address" compared to "ETH withdrawal address" because if RPL withdrawal address isn't set, then RPL goes there too. So calling it just "ETH withdrawal address" is a misnomer.

Agree with all other points. Thanks @Valdorff

darrenlangley:
I have applied your feedback @Valdorff

Valdorff:
> Kane and I discussed this quite a bit. We leaned towards safe-by-default. As you said, you are giving your ETH to the withdrawal address owner. At the time, we couldn't think of a strong use case that demanded not-safe-by-default.

I don't necessarily see this as an either/or. For example, it could be like RPL stake on behalf (allowlisted, with the withdrawal address being allowed by default). I agree open by default is probably not the right path.

Valdorff:
@darrenlangley This looks ready to merge as Draft - let me know if you agree and I'll assign number & merge

Valdorff (approved):
Looks good - confirmed with author it's ready to merge as draft

PR #26, opened 2022-10-01
GMC RPIP

dafuerstman:
RPIP for creation of GMC.

PR #84, opened 2023-10-02
FOR REVIEW: Preamble additions for #26 and #27

LongForWisdom:
Also removed the multiple threads in #26 and replaced them with the summary thread so the portal doesn't break.


xrchz (changes requested):
I think the vote dates should be when the votes conclude, instead of when they started as you have here, based on the (admittedly not very clear) text in RPIP-1.

LongForWisdom:
Dates updated.

xrchz (approved):

PR #90, opened 2023-10-09
FOR REVIEW: Added Obsolete Status to RPIP-1, updated 11, 15, 18.

LongForWisdom:
Also a minor change to the portal to accommodate a larger status field in tables. 
Staging: <https://longforwisdom.github.io/RPIPs/>

**Points to note**
* This requires modification of RPIP-1... Not sure how large or significant modifications you'll allow there absent a vote. 
* Changing language from 'superceded' to 'obsolete'. Obsolete is slightly more generally applicable and we used it at Maker, so  I went with that. If you want to keep it superceded, I can change it back,
* In practice, this prevents the Obsolete RPIPs from appearing on the 'Active' page on the portal. There are present in the 'Home', 'All', and the 'Pages by type' pages.

At one point, I added RPIP-5 to this, given that the response when we asked Val: "Is RPDAO adopted terminology?" Was "What is RPDAO?" After some consideration, I removed this again. RPIP-5 is misleading, but arguably its more misleading marking it as obsolete in its entirety.

### DIFF
--- a/RPIPs/RPIP-1.md
+++ b/RPIPs/RPIP-1.md
@@ -66,7 +66,7 @@ The following is the standardization process for all RPIPs in all tracks:
 
 **Living** - A special status for RPIP that are designed to be continually updated and not reach a state of finality. This includes most notably RPIP-1.
 
-**Obsolete** - An Obsolete RPIP has been replaced, superceded or removed.
+**Obsolete** - An Obsolete RPIP has been replaced, superseded or removed.
 
 Only Final and Living RPIPs are eligible for official adoption.
 
--- a/RPIPs/RPIP-1.md
+++ b/RPIPs/RPIP-1.md
@@ -66,6 +66,8 @@ The following is the standardization process for all RPIPs in all tracks:
 
 **Living** - A special status for RPIP that are designed to be continually updated and not reach a state of finality. This includes most notably RPIP-1.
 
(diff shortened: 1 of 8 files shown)


RPIPs/RPIP-1.md
```diff
@@ -66,6 +66,8 @@ The following is the standardization process for all RPIPs in all tracks:
 
 **Living** - A special status for RPIP that are designed to be continually updated and not reach a state of finality. This includes most notably RPIP-1.
 
+**Obsolete** - An Obsolete RPIP has been replaced, superceded or removed as a result of the passage of another RPIP.
```
Valdorff, line 69: Hmmm... is this meant to only allow for obsolescence due to another RPIP? Wondering both whether it should be left flexible like "...removed, eg, as a result..." or if not whether we should make the language very explicitly ONLY for that case.

LongForWisdom: Uhm. If its been ratified, you definitely don't want to obsolete it without some sort of vote, and currently that's just RPIPs. If its not ratified and its living informational... yeah I can see the argument for this not making sense.

LongForWisdom: I cut down the definition to avoid the issue. Feels like to make it airtight it would be more more specific than the entire rest of RPIP-1, and I would rather tackle it all together at some point in the future.

Valdorff (approved):
Looks good to me

PR #97, opened 2023-10-20
updated after October payments

shfryn:


Valdorff:
image

@shfryn you need to update to latest before you make edits please -- this undoes some govalpha work on making the tables consistent and clear, eg. DM me if you need technical assistance there.

Valdorff (approved):
Looks good - one very minor issue with RA# in titles which is redundant, but not worth holding this up further

PR #98, opened 2023-10-23
RPL withdrawal, ETH on behalf, submission timing feedback

darrenlangley:
In situations, that the RPL provider is different from the node operator:
- An RPL provider should be able to claim RPL rewards
- An RPL provider should be able to stake RPL and not be reliant on the node operator to set them in the whitelist
- `StakeRPL()` should be safe-by-default so that RPL they stake does not go to RPL provider
_This doesn't stop an already staking node operator from setting their RPL withdrawal address to the other provider but we will have a warning in the smart node for this case._

This essentially rounds out the functionality supporting the goal of the RPIP.


RPIPs/RPIP-31.md
```diff
@@ -28,18 +28,45 @@ https://dao.rocketpool.net/t/withdrawal-address-splitter-contract/563
...
+
+### Setting the Address
 - As a node operator, I MUST be able to set an RPL withdrawal address so that I can direct my RPL bond/rewards
   - If a node's RPL withdrawal address is unset, the change MUST come from the node's primary withdrawal address
   - If a node's RPL withdrawal address is set, the change MUST come from the current RPL withdrawal address
 - As a node operator, I MUST be able to confirm the RPL withdrawal address from the new target address before it takes effect, to ensure it is the correct address
 - As the owner of the RPL withdrawal address, I MUST be able to unset my RPL withdrawal address
+
+### Withdrawing Excess RPL
 - As the controller of the RPL withdrawn from a node, I MUST be able to call `RocketNodeStaking.withdrawRPL()` to withdraw excess RPL
```
Valdorff, line 48: In general, I think we should swap these to be independent of implementation. Eg:

> As the controller of the RPL withdrawn from a node, I MUST be able to withdraw excess RPL

or 

> As the controller of the RPL withdrawn from a node, I MUST be able to withdraw excess RPL (eg, by calling `RocketNodeStaking.withdrawRPL()`)

RPIPs/RPIP-31.md
```diff
@@ -28,18 +28,45 @@ https://dao.rocketpool.net/t/withdrawal-address-splitter-contract/563
...
 - If a node's RPL withdrawal address is unset, all RPL rewards and withdrawn RPL will go to the primary withdrawal address
+
+### Setting the Address
 - As a node operator, I MUST be able to set an RPL withdrawal address so that I can direct my RPL bond/rewards
   - If a node's RPL withdrawal address is unset, the change MUST come from the node's primary withdrawal address
   - If a node's RPL withdrawal address is set, the change MUST come from the current RPL withdrawal address
 - As a node operator, I MUST be able to confirm the RPL withdrawal address from the new target address before it takes effect, to ensure it is the correct address
 - As the owner of the RPL withdrawal address, I MUST be able to unset my RPL withdrawal address
+
+### Withdrawing Excess RPL
```
Valdorff, line 47: I'm not sure "excess RPL" is a clear enough concept. I might just have this be "Withdrawing RPL" and then "MUST be able to withdraw RPL as allowed by the protocol" or smth

RPIPs/RPIP-31.md
```diff
@@ -28,18 +28,45 @@ https://dao.rocketpool.net/t/withdrawal-address-splitter-contract/563
...
 - As the controller of the RPL withdrawn from a node, I MUST be able to call `RocketNodeStaking.withdrawRPL()` to withdraw excess RPL
   - If a node's RPL withdrawal address is unset, the call MUST come from one of: the node's primary withdrawal address, or the node's address
   - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address
-- The RPL withdrawal address SHALL be initialized as unset for new node operators
-- The RPL withdrawal address SHALL be initialized as unset for existing node operators
+
+### Claiming RPL rewards
+- As the controller of the RPL for a node, I MUST be able to call `RocketMerkleDistributorMainnet.claim()` to trigger a claim of RPL rewards
+  - If a node's RPL withdrawal address is unset, the call MUST come from one of: the node's primary withdrawal address, or the node's address
+  - If a node's RPL withdrawal address is set, the call MUST come from one of: the node's primary withdrawal address, the current RPL withdrawal address, or the node's address
```
Valdorff, line 55: 🤔 not sure I get this. If the premise is that we have up to 3 totally separate parties - ETH investor, RPL investor, Node Operator - why should all 3 be able to decide when to send RPL rewards out?

I can make a case for the NO, as this is the most "turnkey" approach where the investors need not act at all, but I don't see a case for the ETH investor to have this ability.

RPIPs/RPIP-31.md
```diff
@@ -28,18 +28,45 @@ https://dao.rocketpool.net/t/withdrawal-address-splitter-contract/563
...
 - As the owner of the RPL withdrawal address, I MUST be able to unset my RPL withdrawal address
+
+### Withdrawing Excess RPL
 - As the controller of the RPL withdrawn from a node, I MUST be able to call `RocketNodeStaking.withdrawRPL()` to withdraw excess RPL
   - If a node's RPL withdrawal address is unset, the call MUST come from one of: the node's primary withdrawal address, or the node's address
   - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address
-- The RPL withdrawal address SHALL be initialized as unset for new node operators
-- The RPL withdrawal address SHALL be initialized as unset for existing node operators
+
+### Claiming RPL rewards
```
Valdorff, line 52: Might make sense to have a "Claiming ETH rewards" section for completion here?

RPIPs/RPIP-31.md
```diff
@@ -28,18 +28,45 @@ https://dao.rocketpool.net/t/withdrawal-address-splitter-contract/563
...
+  - If a node's RPL withdrawal address is unset, the call MUST come from one of: the node's primary withdrawal address, or the node's address
+  - If a node's RPL withdrawal address is set, the call MUST come from one of: the node's primary withdrawal address, the current RPL withdrawal address, or the node's address
+- As the controller of the RPL for a node, I MUST be able to call `RocketMerkleDistributorMainnet.claimAndStake()` to trigger a claim of RPL rewards and restake a portion
+  - If a node's RPL withdrawal address is unset, the call MUST come from one of: the node's primary withdrawal address, or the node's address
+  - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address
+
+### Staking RPL
+- As the controller of the RPL for a node, I MUST be able to call `RocketNodeStaking.stakeRPL()` to stake additional RPL
+  - If a node's RPL withdrawal address is unset, the call MUST come from the node's address
+  - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address
```
Valdorff, line 63: This is confusing. The node address is no longer allowed to call stakeRPL here? And the RPL withdrawal address can call _either_ stakeRPL _or_ stakeRPLFor with equivalent results?

RPIPs/RPIP-31.md
```diff
@@ -28,18 +28,45 @@ https://dao.rocketpool.net/t/withdrawal-address-splitter-contract/563
...
+
+### Defaults
+- The RPL withdrawal address SHALL be initialized as unset for new node operators
+- The RPL withdrawal address SHALL be initialized as unset for existing node operators
+  
+### Rewards & Withdrawn RPL
 - If a node's RPL withdrawal address is set, all RPL rewards and withdrawn RPL will go to the RPL withdrawal address
 - If a node's RPL withdrawal address is unset, all RPL rewards and withdrawn RPL will go to the primary withdrawal address
+
+### Setting the Address
```
Valdorff, line 40: Very high level thought about organization... this could also be 3 sections as "Node address powers", "Primary withdrawal address powers", and "RPL withdrawal address powers". I _think_ I like the current structure better, but think it's worth a quick thought

RPIPs/RPIP-31.md
darrenlangley: For this, I have matched what is currently implemented: **withdrawal address AND node address** and added the RPL withdrawal address.

This is so that a node operator cannot withhold rewards and that is it operationally convenient. This would be the same for an ETH investor as well.

RPIPs/RPIP-31.md
darrenlangley: If the RPL withdrawal address is set to a third-party, any RPL staked by the node will be sent to the RPL withdrawal address owner. We would prefer this to be safe-by-default so a node operator cannot fall into this trap. 

I don't really see a case where the node operator would need to stake RPL from the node. If they own the RPL address, it is more convenient to do it from the RPL withdrawal address. If they do not own the RPL address, they are in danger of giving RPL to the withdrawal address owner.

In the draft, I have simply replaced the node operator address with the RPL withdrawal address but we could make that not function inaccessible to the node address, if the RPL withdrawal address is set. That would mean the withdrawal addresses consistently use **stakeRPLFor** instead.

RPIPs/RPIP-31.md
Valdorff: Not tracking - why can the ETH investor force the RPL investor to claim (and not vice versa)?

RPIPs/RPIP-31.md
Valdorff: I _do_ prefer a single way of doing things to two equivalents (as you mention in the last paragraph).

Insofar as why someone would want to avoid using their RPL withdrawal address, for me that's a cold wallet that's (purposely) a little bit of a pain to use. That said, I think this is addressed by `setStakeRPLForAllowed`

I think that would get us to something like:
```
- As the controller of the RPL for a node, I MUST be able to stake additional RPL
  - If a node's RPL withdrawal address is unset, there MUST be a method to do this from the node's address (eg, by calling `RocketNodeStaking.stakeRPL()`), or a specifically allowed address (eg, by calling `RocketNodeStaking.stakeRPLFor()`)
  - If a node's RPL withdrawal address is set, there MUST be a method to do this from the current RPL withdrawal address (eg, by calling `RocketNodeStaking.stakeRPLFor()`), or a specifically allowed address (eg, by calling `RocketNodeStaking.stakeRPLFor()`)
- As the controller of the RPL for a node, I MUST be able to call to control the list of specifically allowed RPL staking addresses (eg, by calling `RocketNodeStaking.setStakeRPLForAllowed()`)
  - If a node's RPL withdrawal address is unset, the call MUST come from the node's address
  - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address
```

RPIPs/RPIP-31.md
darrenlangley: I feel I am missing something as well lol. 

When the RPL address is set, both the ETH investor and RPL investor can trigger a reward claim. I don't dig the fact that smoothing pool rewards and RPL rewards are coupled together but implementation-wise it is much more complicated to separate them.

RPIPs/RPIP-31.md
darrenlangley: hmm not sure what we would put in that section - are there relevant items to be addressed for ETH claiming?

RPIPs/RPIP-31.md
darrenlangley: The only issue with the "there MUST be a method to do this" wording is that it doesn't explicitly say that the node operator cannot stake RPL. I think we can remove the implementation details and it makes the spec clearer. 
How is this?
```
- As the controller of the RPL for a node, I MUST be able to stake additional RPL
  - If a node's RPL withdrawal address is unset, the call MUST come from the node's address, or a specifically allowed address
  - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address, or a specifically allowed address
- As the controller of the RPL for a node, I MUST be able to control the list of specifically allowed RPL staking addresses
  - If a node's RPL withdrawal address is unset, the call MUST come from the node's address
  - If a node's RPL withdrawal address is set, the call MUST come from the current RPL withdrawal address
  ```

File: RPIPs/RPIP-31.md (hunk 4 above)
Comment (Valdorff) [reply RPIPs/RPIP-31.md:52]: Yes! The thing I was missing from above!

Namely that the RPL investor can force an ETH distribution if in smoothing pool

File: RPIPs/RPIP-31.md (hunk 3 above)
Comment (Valdorff) [reply RPIPs/RPIP-31.md:55]: Got it, it's cuz of the smoothing pool. If not in smoothing pool does that still apply?

File: RPIPs/RPIP-31.md (hunk 5 above)
Comment (Valdorff) [reply RPIPs/RPIP-31.md]: Lgtm

File: RPIPs/RPIP-balanceprice-timing.md [hunk 7]
```diff
@@ -0,0 +1,39 @@
...
+
+*(As a side-node, this also makes the implementation of the watchtower easier and more forgiving (computation-wise) on the Execution Client; balances are calculated by "approximating" the rewards distribution of the Smoothing Pool at each balances report, and this change means balance reporting can use the same network snapshot as rewards calculation so the watchtower won't need to generate the state twice for the same slot.)*
+
+## Specification
+
+- As the oDAO watchtower, I MUST be able to read the next time-based balance submission interval from the Rocket Pool protocol
+- As the oDAO watchtower, I MUST be able to read the next time-based RPL price submission interval from the Rocket Pool protocol  
+
+## Backwards Compatibility
+There is a slim chance that DeFi protocols that rely on the rETH exchange rate calculation may be affected, as the timing will change. That said, it is unlikely that DeFi protocols are reliant on the specific timing. Going forward, it makes DeFi integrataions more intuitive because the rETH exchange rate changes more consistently.
```
Valdorff, line 33: Typo integrations

RPIPs/RPIP-balanceprice-timing.md
```diff
@@ -0,0 +1,39 @@
...
+
+This change removes the uncertainty associated with this timing by shifting price and balance intervals from a number based on blocks to a number based on seconds, which has parity with the way rewards intervals are calculated today. Essentially it means prices and balances will be reported at the same time each day (assuming the interval is set to `86400`, the number of seconds in 24 hours) without exception. It standardizes the timing of both functions to make them easier to predict.
+
+Furthermore, the coupling of the rewards period to this interval means that the "top up window" will now be a fixed quantity directly controllable by the protocol; it will start at the beginning of the final balances reporting interval of the rewards period, and end with the rewards period end. Thus, if the balances period is set to 1 day, users will consistently have exactly 24 hours to top up from the final price update of the interval. The start and end times can be reliably provided so the guess-work is removed.
```
Valdorff, line 23: Should spec include a starting value for this setting (eg one day)?

darrenlangley: The idea is that it will be set to one day but we can make that explicit in the spec.

RPIPs/RPIP-31.md
darrenlangley: Currently, the node address and primary withdrawal address can call `RocketMerkleDistributorMainnet.claim()` regardless of whether they are in the smoothing pool or not.

We could revert node and primary withdrawal address calls if they are not in the smoothing pool but it then creates an inconsistent experience for regular users that use the RPL withdrawal address - as soon as they join the smoothing pool they can no longer claim from their node or primary withdrawal address.

darrenlangley:
@xrchz had some feedback on the staking ETH on behalf of feature:
https://dao.rocketpool.net/t/rpip-32-stake-eth-on-behalf-of-node/2252

Valdorff (approved):
LGTM. Author requested merge in DM.

PR #100, opened 2023-10-30
DRAFT: More Latex 'fixes'

LongForWisdom:
Very hacky fix to unify latex parsing between portal and github. 

This is really dumb, but there is at least now some valid input that results in the same output on both platforms, which was not previously the case. 

Is this a massive hack to get the portal working? Absolutely yes. 

Does it change other aspects of the portal (specifically how code blocks are rendered)? Also yes. (Though I think its bearable, or I wouldn't be making the PR.)

Do I recommend taking this PR? I guess. Not excited about it. It does render better now, though.

### Change to Latex Rendering

**Before**

image

**After**

image

**Github Render**

image

**Input**

image

### Collateral Damage to Code Blocks

**Before**

image

**After (border gone)**

image


LongForWisdom:
Hold off til tomorrow. I managed to do the pges upgrade which was way simpler than I thought. And I want to see if it just fixes the latex issues completely, without needing any of this jank.

PR #102, opened 2023-11-01
FOR REVIEW: More Minor RPIP Fixes

LongForWisdom:
Very minor fixes that can just editor-merge.
* Fixing RPIP template link in RPIP-1
* Remove reference to RFC822, it just confuses things if you're 99% of people that don't know what this is. Also not 100% sure 822 is even the right number.
* Add hyperlinks to RPIP10 rationale.
* Change absolute links for template to point to main rather than master
* Fixed template to not break due to numerical sort if left unchanged in draft RPIP.

I've tested locally and all seems good, but I've not put it up on staging because this is minor and I'm using my staging branch for other stuff atm.

### DIFF
--- a/RPIPs/RPIP-1.md
+++ b/RPIPs/RPIP-1.md
@@ -80,7 +80,7 @@ RPIPs should be written in markdown format.
 
 **RPIP Header Preamble**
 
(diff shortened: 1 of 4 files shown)


Valdorff (approved):
No material changes - just pure url best practices, clarifications, and in one place a missing word

PR #103, opened 2023-11-01
FOR REVIEW: Removed Review status. 

LongForWisdom:
Also added a reference to Obsolete I missed in that change (which I'll do separately if this is rejected.)

Based on discussions you may not want this as is. If so, just close, no big deal.


Valdorff:
Right... I think we mostly need to use Review earlier (and I put myself foremost on that list). I actually thought "Review" had more encumbering process than it does, but not really. I'd like to add a block diagram and a few words to try to clarify a useful flow.

PR #112, opened 2023-11-10
Updated disband and RPL locking permissions

darrenlangley:
**Removed disband** - pDAO can still disband the security council by removing all members.
**RPL locking permissions** - provides a way for node operators to opt into governance processes and also in situations where RPL does not belong to the node operator it prevents node operators from maliciously proposing and challenging to steal or grief RPL providers.

### DIFF
--- a/RPIPs/RPIP-33.md
+++ b/RPIPs/RPIP-33.md
@@ -146,7 +146,9 @@ The purpose of the proposal type is to control the set of members of the "Securi
 possible:
 
 1. **Invite Member:** Invites a new member to the Security Council.
-2. **Remove Member:** Removes an existing member from the Security Council.
+2. **Remove Members:** Removes existing members from the Security Council.
+
+Note that the security council MAY be effectively disabled by removing all members.
 
 The pDAO also has control over how many members are required for quorum. This is adjusted with a parameter change of the
 parameter `security.members.quorum`.
--- a/RPIPs/RPIP-33.md
+++ b/RPIPs/RPIP-33.md
@@ -147,7 +147,6 @@ possible:
 
 1. **Invite Member:** Invites a new member to the Security Council.
 2. **Remove Member:** Removes an existing member from the Security Council.
-3. **Disband:** Removes all members from the Security Council. Effectively disabling it.
 
 The pDAO also has control over how many members are required for quorum. This is adjusted with a parameter change of the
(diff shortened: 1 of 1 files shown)


RPIPs/RPIP-33.md
```diff
@@ -198,6 +197,8 @@ for the duration of the proposal process. In order to be eligible to propose, no
 for the purposes of rewards, voting and collateral requirements. Locked RPL SHALL NOT be counted towards thresholds for
 withdrawing RPL.
 
+By default, locking RPL for any purpose will be disabled. Node operator's will opt-in to performing governance activities by enabling the locking of RPL from their node or primary withdrawal address. On the condition that [RPIP-31: RPL Withdrawal Address](./RPIP-31.md) is ratified and implemented, if the RPL withdrawal address is set then, you will only be able to enable or disable RPL locking using this address.
```
Valdorff, line 200: `locking RPL for any purpose` is too strong I think. Maybe "for any purpose defined in this RPIP" or enumerate them? Just don't want to accidentally have this interact with future proposals.

Valdorff:
Can't put this on the appropriate line cuz unmodified, but `proposal.security.quorum` isn't in the table and `members.quorum` is. These should match.

RPIPs/RPIP-33.md
darrenlangley: yeah good point. At this stage, the functionality is quite broad but from an RPIP perspective we should only have it in relation to this RPIP and include any future behaviour in future RPIPs.

I will update.

RPIPs/RPIP-33.md
```diff
@@ -147,7 +147,6 @@ possible:
 
 1. **Invite Member:** Invites a new member to the Security Council.
 2. **Remove Member:** Removes an existing member from the Security Council.
-3. **Disband:** Removes all members from the Security Council. Effectively disabling it.
```
Valdorff, line 150: If we're removing this, the above should allow for removing multiple members at once (else we have a period where the security council is just one member as we remove folks one at a time). I would also suggest saying something like "Note that the security council MAY be effectively disabled by removing all members."

darrenlangley: No problem, Kane is going to adjust the remove function to be able to remove multiple members.

Valdorff (approved):
LGTM

PR #113, opened 2023-11-10
Assign RPIP number to balance-price-timing

LongForWisdom:
Per RPIP-1, number assigned and date set to date of number assignment.


Valdorff (approved):
Good catch - should have a number to be merged here.

Valdorff:
@darrenlangley could you add a forum thread to this? If not, I'll merge in a few hours anyhow, but would save the later step if you can get to it.

darrenlangley:
> @darrenlangley could you add a forum thread to this? If not, I'll merge in a few hours anyhow, but would save the later step if you can get to it.

Actually, I think this is chicken and egg. It would be good to have an RPIP number for the forum post so that it is recognisable later. Would be great if you can merge and I will make an update for the forum post. I have the forum post drafted.

PR #114, opened 2023-11-13
FOR REVIEW: Portal Improvements - Updates, LaTeX, etc

LongForWisdom:
Ended up being more than just latex stuff. This PR includes.

**Major Changes**
* Updating to most recent github pages version. 
* Modified parser from Kramdown to GFM. This should fix the issues we've seen with tables.
* Updated to MathJax 3 (LaTeX renderer.)

**Minor Changes**
* Updated the readme to reflect that there is an automatic build process now. 
* Fixed VOPs image links
* Fixed RPIP-30 image links
* Fixed RPIP-30 LaTeX output. See [staging](https://longforwisdom.github.io/RPIPs/RPIPs/RPIP-30#specification).
* Fixed RPIP-33 LaTeX output. See [staging](https://longforwisdom.github.io/RPIPs/RPIPs/RPIP-33#voting-power-function).
* Added CSS to fix block LaTeX display. 
* Added 'how-to-LaTeX' instructions to the RPIP template.

**Caveats**
* Inline LaTeX is hit or miss. This is the best we can get without a non-trivial time investment.
* LaTeX rendering on firefox looks janky, but is fully readable. This is due to use of the :has pseudo-class. My understanding is that support for this should make it into firefox at some point in the next year or so.

At this point I'm fairly happy with the changes and results, and would recommend taking the PR.


Valdorff (approved):
This looks good to me both skimming code and looking at staging.

Thanks for the effort on this esoteric bit of fun!

PR #117, opened 2023-11-28
Rpl withdrawal

darrenlangley:
Based on some devnet testing we have the following adjustment.

**Staking RPL**
The current behaviour is that the primary withdrawal address can stake RPL. The RPIP's unset condition excluded the primary withdrawal address - this would make it not backwards compatible. I suspect quite a few people use that feature.


Valdorff (approved):
LGTM

PR #119, opened 2023-11-28
FOR REVIEW: Updated RPIP-28 and RPIP-30 with vote results.

LongForWisdom:
https://longforwisdom.github.io/RPIPs/active

image
image
image

Note per previous convention, have gone with 2023-11-16 and not worried too much about the TZ difference (shows as 17th here)


htimsk (approved):
Looks good.

PR #120, opened 2023-11-28
updated with Nov awards

shfryn:


Valdorff (approved):
Note the dates of awards have been changed to reflect when the challenge period expires.

Looks good. I hate reviewing tables, and am glad there's decent diff tools out there to help highlight things.

PR #131, opened 2024-01-25
Updated expense label with correct dev funding period

darrenlangley:


Valdorff (approved):
This had previously been fixed elsewhere. Good catch.

PR #134, opened 2024-02-06
Add RPIP-38

jshufro:


Valdorff (approved):
Accelerating to review. Vote text and RPIP should both get reviewed before this is finalized.

PR #151, opened 2024-03-06
Now complete draft for committee stipends.

LongForWisdom:


RPIPs/RPIP-committee_stipends.md
```diff
@@ -0,0 +1,130 @@
...
+$$
+```
+
+#### Parameter Changes
+
+* The Target Hours input for a given committee MAY be adjusted via the internal vote of that committee. 
+  * Increases to Target Hours MUST have a unanimous positive vote outcome, which MUST be posted publicly.
+  * Decreases to Target Hours MUST have a majority positive vote outcome, which MUST be posted publicly. 
+  * Any Target Hours change is subject to pDAO challenge, and SHALL NOT be acted upon until 1 calendar month has passed after the decision has been announced.
+* The Target Member Count SHALL be modified only by the pDAO via a voted modification to this RPIP and SHOULD follow the usual processes for such a modification.
```
Valdorff, line 55: Hmmm... 

For the GMC, https://rpips.rocketpool.net/RPIPs/RPIP-40 has `The GMC SHALL contain nine individuals, except in cases of vacancies.`

For the IMC, there is no such limitation. It's simply the size per the selection RPIP (though the IMC could boot folks with a 2/3 majority). RPIP-10 does establisht that all MCs are at least a 3-of-5.

I think having selection vote set this might be the move?

RPIPs/RPIP-committee_stipends.md
```diff
@@ -0,0 +1,130 @@
...
+3. Replace this link with your updated sheet
+4. Convert the new sheet to markdown using: https://www.tablesgenerator.com/markdown_tables or similar.
+-->
+
+### Payments
+
+#### Split
+The split of the committee stipend budget SHOULD be determined by the voting members of the committee in question. 
+* Each committee SHOULD be consistent in the method used to determine the payout split. 
+* Roles that are compensated through other mechanisms SHOULD be discounted when the committee determines the stipend split.
```
Valdorff, line 91: And the target member hours?

RPIPs/RPIP-committee_stipends.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Distribution
+The Grants Management Committee is responsible for distributing the stipends for each committee directly to its members.
+* Stipends SHOULD be paid out in one of the GMC's accepted payout token options, selected by the recipient.
+  * Conversion from USD to dollar-stablecoin payout token SHOULD be priced 1:1. 
+  * Conversion from USD to any other payout token SHOULD be done using the market price at the time of payment transaction creation. 
+* Other committees MUST provide the GMC Administrator with the details of their stipend split each month. 
+  * They MUST communicate payment addresses, stipend share in USD, and requested payment token for each of their members. 
+  * They MUST meet any deadline or format requirements imposed by the GMC Administrator.
+* Any undistributed funds for a given month are forfeited by that committee.
```
Valdorff, line 101: This seems un-necessarily punitive. If it becomes a problem, I'm fine with adding it, but if I currently had to guess why such a thing would happen it's like "people being busy with a crypto conference" or something. Ie, we might piss people off or make them choose to attend to the urgent over the important.

LongForWisdom: So, I think this is actually fairly important. If there are not hard deadlines + consequences for missing them, then the burden to 'catch-up' ends up on the GMC admin. I had a lot of experience at this at Maker, and it sucks to be in those shoes and have to decide whether to be punitive to others or just accept the cost to yourself. 

Given that the deadline is set by the GMC admin, they can adjust said deadline if there is a conference or such coming up. I'd expect them to provide common courtesy and adjust the deadline if it was generally agreed to be necessary.

RPIPs/RPIP-committee_stipends.md
Valdorff: > Might be simpler to just return it to current # of members.
This sounds like the move as it avoids the interactions with MC charters

RPIPs/RPIP-committee_stipends.md
Valdorff: Hmmm... fair... the fact that it's a human doing the imposing should provide a lot of wiggle room. Convinced enough.

RPIPs/RPIP-committee_stipends.md
Valdorff: I don't see that covered in the section above. Eg, if we pay out to "Treasurer" in a new way, that work should no longer be included when calculating TargetHours.

LongForWisdom: Right, I misunderstood you initially. Agree, I'll make the change.

Valdorff (approved):
Looks good and thanks for addressing my comments

PR #160, opened 2024-03-26
Forced exits for offline nodes via keeper mechanism

VVander:
This is a draft proposal which outlines one way to use EIP-7002 to improve Rocket Pool. Specifically, I propose a keeper mechanism to incentivize all users to keep offline nodes accountable.

### DIFF
--- a/RPIP-forced-exits.md
+++ b/RPIP-forced-exits.md
@@ -1,0 +1,67 @@
+---
+rpip: Unassigned
+title: Forcibly Exit Offline Validators
+description: Use EIP-7002 to exit validators which are offline for a significant period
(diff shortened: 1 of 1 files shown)


Valdorff:
This seems to be a more-fleshed out subset of RPIP-44. Is there a particular benefit to breaking this out, or should some of the fleshing out be ported instead?

VVander:
I wasn't sure what was happening with RPIP-44 since it doesn't specify a mechanism, so I just wrote down my complete idea here. If you want to re-assign RPIP-44 to point to this doc that's ok with me, or you can always keep the other one if it's going in a different direction.

PR #166, opened 2024-04-05
Added vote data to all the recent RPIPs. Updated portal to handle multiple votes in a minimum-viable sort of manner.

LongForWisdom:
Staging for testing here: <https://longforwisdom.github.io/RPIPs/>


Valdorff (approved):
double checked all dates and results. checked first few characters in vote-to proposal urls

PR #168, opened 2024-04-09
Add Rocket Rewards Tree Informational RPIP

xrchz:


RPIPs/rpip-draft_rewards_tree.md
```diff
@@ -0,0 +1,36 @@
+---
+rpip: #<to be assigned>
+title: Rewards Tree Specification
```
Valdorff, line 3: Maybe "Current Rewards Tree Specification"

RPIPs/rpip-draft_rewards_tree.md
```diff
@@ -0,0 +1,36 @@
...
+status: Draft
+type: Informational
+created: 2024-04-09
+---
+
+## Abstract
+This RPIP describes the current state of the specification of the Rocket Pool rewards system (first introduced in [Redstone](https://medium.com/rocket-pool/rocket-pool-the-merge-redstone-601d9efd6b4)) for node operators, wherein ETH and RPL rewards are allocated at regular intervals to staking nodes.
+
+## Motivation
+Rocket Pool's allocation of rewards to node operators needs a clear specification that can be independently implemented and verified so that node operators know what to expect and have a basis for recourse in case of any misallocation.
```
Valdorff, line 16: "and have a basis for recourse in case of any misallocation" is not currently implied anywhere and pretty significant. I'd suggest simply removing or maybe "know what to expect and are able to verify allocations." or something.

RPIPs/rpip-draft_rewards_tree.md
```diff
@@ -0,0 +1,36 @@
...
+
+The rewards tree specification is currently developed in the external [Rocket Pool Research](https://github.com/rocket-pool/rocketpool-research/tree/master/Merkle%20Rewards%20System) repository. Development there began before the RPIP process was created. However, since this is a core protocol specification, its proper home is as an RPIP where it can be easily found and referred to with a clear status as an official definition of (part of) the Rocket Pool protocol. This RPIP is an informational document describing the status quo, with the intention of moving the core protocol specification into a future RPIP.
```
Valdorff, line 18: "with the intention of following up with a future pDAO-ratified RPIP (which may or may not include changes)."

RPIPs/rpip-draft_rewards_tree.md
```diff
@@ -0,0 +1,36 @@
...
+* [Rewards Calculation Specification](../assets/rpip-tree/rewards-calculation-spec.md)
+
+## Rationale
+This specification describes what is actually implemented and used by Rocket Pool as of April 2024 (that is, for rewards interval 21).
+
+## Reference Implementation
+This specification (Rewards Calculation v8) is implemented by [Treegen](https://github.com/rocket-pool/treegen/tree/a92de2f75e8188c450653b01af5ab27f97a14ec3), which is the reference implementation. It is also implemented by [Rockettree](https://github.com/xrchz/rockettree/tree/ce9bca3d15ce490804ea8e31ebfd5a24723dc8cf).
+
+## Security Considerations
+Errors, inconsistencies, and omissions in this specification leave the protocol in a difficult position should there arise any disputes over the allocation of rewards.
```
Valdorff, line 33: Again - I don't think we should be implying recourse where that (a) has not been implied previously, (b) has no suggested implementation and (c) has not been ratified. I don't really think you need this "Security Considerations" section for this "what's the state" info RPIP.

RPIPs/rpip-draft_rewards_tree.md
```diff
@@ -0,0 +1,36 @@
...
+The technical specification (v8 in particular) taken from the external [research repository](https://github.com/rocket-pool/rocketpool-research/tree/master/Merkle%20Rewards%20System) is attached as the following assets:
+
+* [Merkle Tree Specification](../assets/rpip-tree/merkle-tree-spec.md)
+* [Rewards Calculation Specification](../assets/rpip-tree/rewards-calculation-spec.md)
+
+## Rationale
+This specification describes what is actually implemented and used by Rocket Pool as of April 2024 (that is, for rewards interval 21).
+
+## Reference Implementation
+This specification (Rewards Calculation v8) is implemented by [Treegen](https://github.com/rocket-pool/treegen/tree/a92de2f75e8188c450653b01af5ab27f97a14ec3), which is the reference implementation. It is also implemented by [Rockettree](https://github.com/xrchz/rockettree/tree/ce9bca3d15ce490804ea8e31ebfd5a24723dc8cf).
```
Valdorff, line 30: Does the watchtower import treegen, or does it have a separate copy? If the latter, it's worth linking here too I think. If the former, worth noting.

Valdorff (commented):
A few thoughts

Valdorff:
Confirmed that the assets are the same. `rewards-calculation-spec.md` has 2 links changed from relative to absolute external. The external links point to a specific commit and are non-essential. While they may conceivably get broken by external repo actions, it won't mislead users.

RPIPs/rpip-draft_rewards_tree.md
xrchz: For the future Core Protocol version I was hoping the legitimacy from being a ratified RPIP would be the basis for recourse. Are you objecting to that, or just to it being present in the Informational version?

RPIPs/rpip-draft_rewards_tree.md
xrchz: Done

RPIPs/rpip-draft_rewards_tree.md
xrchz: Done

RPIPs/rpip-draft_rewards_tree.md
xrchz: Removed the section (it was there from when this was gonna be Protocol).

RPIPs/rpip-draft_rewards_tree.md
xrchz: So... no change required here, if I understand correctly.

jshufro: I'd suggest just linking to `https://github.com/rocket-pool/treegen`

The permalinks are nice, but confusing if they become outdated.

I don't think you need to separately link watchtower code. The treegen repo is the defacto starting place for anyone who wants to learn more.

RPIPs/rpip-draft_rewards_tree.md
xrchz: (after chat elsewhere) rewording as suggested

Valdorff (approved):
Looks good to me. Assigned number and set status to Final upon author request.

PR #170, opened 2024-04-10
RPIP-41: Committee Stipends - Request Finalization prior to vote

LongForWisdom:


Valdorff (approved):
Author requests Final status. Text has gotten feedback. Sentiment poll looks solid.

PR #172, opened 2024-04-11
RPIP-41 - Explicitly state that stipends are funded from the GMC's budget.

LongForWisdom:


Valdorff:
Author was asked a question in the vote text thread. While the general assumption is that the GMC spends the GMC's money, this PR makes it explicit (woo explicit). Myself and 1 other RPIP editor have agreed that this is clarifying what's already there, which means it's an appropriate edit despite already being Final.

Valdorff (approved):
Useful clarification, see previous comment

PR #175, opened 2024-04-18
RPIP-41 - Very minor clarification changes to resolve some questions raised in the vote text discussion.

LongForWisdom:


Valdorff (approved):
I agree that these are all further clarifying things that are already unambiguous. LGTM

PR #174, opened 2024-04-16
RPIP-52: Rewards Specification v9

jshufro:

PR #176, opened 2024-04-19
Added RPIP-53 (Rewards v10)

jclapis:
This is a PR to add RPIP-53 to the collection, which covers a proposal for rewards spec v10. I called it that because even though [Patches's v9 proposal](https://github.com/rocket-pool/RPIPs/pull/174) is still a draft, his came first and this isn't mutually exclusive with it.

### DIFF
--- a/RPIPs/RPIP-53.md
+++ b/RPIPs/RPIP-53.md
@@ -1,0 +1,110 @@
+---
+rpip: 53
+title: Rewards Tree v10 Specification
+description: Adjusts the rewards file to account for Houston changes, generalizing rewards from nodes to claimers and supporting the RPL withdrawal address
+author: Joe Clapis (@jcrtp)
+discussions-to: TBD
+status: Draft
+type: Protocol
+created: 2024-04-19
+---
+
+
+## Abstract
+
+The incoming Houston upgrade introduces [RPIP-31](https://github.com/rocket-pool/RPIPs/blob/main/RPIPs/RPIP-31.md) - the **RPL withdrawal address**. Unfortunately, node operators using this new feature are **no longer able to claim rewards** if they have an RPL withdrawal address. *Only* the RPL withdrawal address is able to claim their rewards. This includes both RPL *and* Smoothing Pool ETH rewards.
+
+This RPIP proposes to generalize the rewards tree, changing *nodes* to *claimers*. The tree will now specify the amount of RPL and Smoothing Pool ETH available to individual claimers, where:
+- If a node does not have an RPL withdrawal address set, their node address is the claimer address (same as today).
+- If a node *does* have an RPL withdrawal address set, the RPL withdrawal address will be the claimer address. The node's earnings will be added to the claimer's total rewards. The node itself will no longer appear in the rewards tree.   
+
+This will improve rewards claiming UX, remove some inconsistencies between the existing ruleset and this new feature, make the rewards system more attractive to users that would actually leverage RPL withdrawal addresses, and will not adversely impact nodes that do not use it.
+
+This **does not** require a contract change (ironically, it's more in-line with how the contracts actually behave in Houston) and can be implemented at any time after Houston has launched.
+
+
+## Motivation
+
+The Redstone update introduced the Merkle-tree-based rewards system, which allowed users to claim RPL they've earned for both node operating duties and Oracle DAO duties, as well as their share of Smoothing Pool rewards. Since its inception, this system has gone through several revisions to account for fixes and Ethereum upgrades but has fundamentally remained the same: the tree lists rewards earned for individual node operators, and node operators can claim those rewards at any time.
+
+On that note, Houston (or more specifically RPIP-31) has an unfortunate consequence: if you have an RPL withdrawal address set, [you can no longer claim rewards intervals](https://github.com/rocket-pool/rocketpool/blob/a747457fc610aa889af0a13cff5d3d00955525a5/contracts/contract/rewards/RocketMerkleDistributorMainnet.sol#L85). This includes both RPL rewards and Smoothing Pool ETH rewards. Claiming them *must* be done from the RPL withdrawal address. Whether intended or not, the RPL withdrawal address isn't really an RPL withdrawal address in Houston; it's a "claimer address" that can claim rewards intervals from the Merkle distributor contract.
+
+This is problematic for a few reasons. As an example: the node is still going to be listed in the rewards file, and thus the rewards for the corresponding interval are still going to show up as a "claimable interval" in the Smart Node despite the user not being able to claim from it, and not being able to actually *receive* any of the rewards from that interval. It will only disappear from the list once the RPL withdrawal address has claimed from it. This can obviously be worked around in the Smart Node but it's exactly that: a workaround. The file itself is still going to have the node operator listed, so anyone (such as a third party) that wants to use it to view or claim rewards has to know about and implement this workaround as well.
+
+In this RPIP, I'm proposing a small change to the rewards file that ameliorates the problems and provides some helpful benefits for people that would actually take advantage of the feature in the first place.
+
+
+## Important Clarification
+
+Note that the RPL withdrawal address, as specified in RPIP-31, is **not** the same as the "primary withdrawal address" (previously known as just the "withdrawal address" in Atlas and below). If a node operator has a "regular" withdrawal address set, they will *not be affected by this change*. They can still claim rewards normally. The only nodes affected by this change are ones that explicitly set an **RPL withdrawal address**, which is a different setting than the "primary withdrawal address", after Houston has been deployed.
+
+
+## Specification
+
+The specification is going to be one of two possibilities:
+- Either an amendment to v8 [from the research repo]((https://github.com/rocket-pool/rocketpool-research/blob/master/Merkle%20Rewards%20System/merkle-tree-spec.md)), or
+- An amendment to v9 specified in [candidate RPIP-52](https://github.com/rocket-pool/RPIPs/pull/174) by Patches. 
+
+This change itself is orthogonal to the decision of whether or not v9 is ratified, as v9 is a non-functional format change from JSON to SSZ, but the specification *itself* does depend on whether or not v9 is ratified so it can build upon the proper baseline. I'm calling this "v10" because I anticipate v9 will go through and come first. I'm happy to formalize the specification once the decision has been made. In the mean time, I'll propose the changes here at a high-level.
+
+Alternatively I can adjust both the v8 and v9 specs accordingly, just so we have them ready to go. I defer to the community to decide whether or not that's prudent. Anyway, high-level changes for now:
+
+
+### Header Changes
+
+- The `RewardsFileVersion` will change from 3 to 4.
+- The `RulesetVersion` will change from 8 to 9 (or 10 if RPIP-52 is ratified first).
+
+
+### Body Changes
+
+The `node_rewards` section of the rewards file, which lists the rewards earned by individual node operators, will be replaced by a more generic `claimer_rewards` section. `claimer_rewards` lists the rewards available to be claimed for the interval *by a claimer*, which may or may not be a node operator.
+
+The structures of `node_rewards` and `claimer_rewards` are identical. It's just a name change; purely a semantic difference. The rules for how elements are added to that list, however, are new.
+
+A claimer (an entry in the `claimer_rewards` list) is determined as follows:
+- Calculate the earnings (RPL and Smoothing Pool) of each node in the network normally.
+- For each node, determine the `claimer` address:
+  - If the node does *not* have an RPL withdrawal address set at the `targetSlot` of the interval (the slot at which rewards are calculated), the claimer is the node's address. This is the same as it is today, so if a node doesn't use the RPL withdrawal address feature then its experience is unchanged.
+  - If a node *does* have an RPL withdrawal address set at the `targetSlot`, the claimer is the node's RPL withdrawal address.
+- Add the node's rewards to the entry in `claimer_rewards` for the `claimer` address. If one does not exist, it is created. Note there is **one entry per claimer**.
+    - *(In other words, the claimer represents the aggregated sum of all rewards earned during the interval by either that node operator, or by node operators that have set their RPL withdrawal address to that particular address.)*
+
+To be clear, "whether or not an RPL withdrawal address is set" looks specifically at the value returned by the [`RocketNodeManager.getNodeRPLWithdrawalAddressIsSet`](https://github.com/rocket-pool/rocketpool/blob/a747457fc610aa889af0a13cff5d3d00955525a5/contracts/contract/node/RocketNodeManager.sol#L121) contract call. It doesn't compare the RPL withdrawal address with the node address or the primary withdrawal address; it simply keys off of the boolean returned by that view.
+
+
+## Rationale
+
+This change has some obvious benefits:
+
+1. Nodes without an RPL withdrawal address set at the time of `targetSlot` are unaffected by this change.
+2. Nodes that earned rewards during an interval, but have an RPL withdrawal address set, will not be present in the file. This naturally makes the Smart Node interpret that as being "ineligible to claim" for the interval and will thus hide it from the node operator in the list of intervals presented when claiming rewards.
+3. The RPL withdrawal address is largely targeted at entities that would take advantage of a splitter contract (e.g., whale marriages) or entities building on top of Rocket Pool (e.g., NodeSet). In the situation where multiple nodes have the same RPL withdrawal address assigned, this reduces the number of transactions required for that claimer to claim rewards from one-per-node to just one. It reduces O(n) gas costs down to O(1).
+4. This **does not** require any contract changes. It works with the Houston variant of [`RocketMerkleDistributorMainnet`](https://github.com/rocket-pool/rocketpool/blob/a747457fc610aa889af0a13cff5d3d00955525a5/contracts/contract/rewards/RocketMerkleDistributorMainnet.sol) (which is where it would apply).
+   1. The contract doesn't actually check if the claimer is a registered node or not, despite what you might think. It just checks if the claimer has an RPL withdrawal address set or not, which is why this whole thing works in the first place.
+   2. Ostensibly, this behavior is more in-line with how the contracts actually function than the v8 / v9 system.
+5. This will not increase the size of the rewards file. In the "worst case", where *nobody* has an RPL withdrawal address set, the size is the same as it would be without this change. In the "best case", where *everybody* has the RPL withdrawal address set to the same address, the tree would only consist of one entry.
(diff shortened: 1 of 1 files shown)


RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Important Clarification
+
+Note that the RPL withdrawal address, as specified in RPIP-31, is **not** the same as the "primary withdrawal address" (previously known as just the "withdrawal address" in Atlas and below). If a node operator has a "regular" withdrawal address set, they will *not be affected by this change*. They can still claim rewards normally. The only nodes affected by this change are ones that explicitly set an **RPL withdrawal address**, which is a different setting than the "primary withdrawal address", after Houston has been deployed.
+
+
+## Specification
+
+The specification is going to be one of two possibilities:
+- Either an amendment to v8 [from the research repo]((https://github.com/rocket-pool/rocketpool-research/blob/master/Merkle%20Rewards%20System/merkle-tree-spec.md)), or
```
xrchz, line 45: This should presumably refer to RPIP-51 rather than externally.

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,110 @@
...
+
+In this RPIP, I'm proposing a small change to the rewards file that ameliorates the problems and provides some helpful benefits for people that would actually take advantage of the feature in the first place.
+
+
+## Important Clarification
+
+Note that the RPL withdrawal address, as specified in RPIP-31, is **not** the same as the "primary withdrawal address" (previously known as just the "withdrawal address" in Atlas and below). If a node operator has a "regular" withdrawal address set, they will *not be affected by this change*. They can still claim rewards normally. The only nodes affected by this change are ones that explicitly set an **RPL withdrawal address**, which is a different setting than the "primary withdrawal address", after Houston has been deployed.
+
+
+## Specification
```
xrchz, line 42: I'm not happy for a Protocol RPIP to have its Specification given as an amendment only. I think it's good to include the amendment in the RPIP, but for the Specification itself to be a full copy of the specification (probably as an asset attached, as with the previous RPIPs specifying rewards).

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,110 @@
...
+   2. Ostensibly, this behavior is more in-line with how the contracts actually function than the v8 / v9 system.
+5. This will not increase the size of the rewards file. In the "worst case", where *nobody* has an RPL withdrawal address set, the size is the same as it would be without this change. In the "best case", where *everybody* has the RPL withdrawal address set to the same address, the tree would only consist of one entry.
+6. The implementation is actually quite simple. I was able to put the reference implementation together in like half an hour. I wouldn't have any contention about making this the first ruleset after Houston launches once it's cleared the usual cross-implementation testing gamut.
+
+With that being said, there are some important implications that need to be considered:
+
+1. This effectively "locks in" the rewards for an interval to a particular claimer. If you have an RPL withdrawal address set at the time the interval is created, and the address later changes your node's RPL withdrawal address to something else, the new address won't be able to claim the rewards. Only the original RPL withdrawal address will.
```
xrchz, line 91: This is the opposite behaviour for changing the normal withdrawal address... it seems pretty bad/unintuitive...

RPIPs/RPIP-53.md
jclapis: Yep, you're right - I will amend it to link to RPIP-51 instead. Thanks for the catch.

RPIPs/RPIP-53.md
jclapis: As stated in the Specification section:

> This change itself is orthogonal to the decision of whether or not v9 is ratified, as v9 is a non-functional format change from JSON to SSZ, but the specification itself does depend on whether or not v9 is ratified so it can build upon the proper baseline. I'm calling this "v10" because I anticipate v9 will go through and come first. I'm happy to formalize the specification once the decision has been made. In the mean time, I'll propose the changes here at a high-level.
> 
> Alternatively I can adjust both the v8 and v9 specs accordingly, just so we have them ready to go. I defer to the community to decide whether or not that's prudent. Anyway, high-level changes for now:

I was just waiting to see whether or not v9 would be formalized first, since the spec changes based on whether or not it is and it's still in flight. Sounds like you're lobbying for the "do both specs and we'll pick whichever one is relevant at the time" camp, which is fine; it'll just take a little longer to write and I wanted to get this out before that.

RPIPs/RPIP-53.md
jclapis: Depends; there are people that think this is actually an improvement over the current system to prevent exploited addresses from claiming everything. I tend to lean in that camp personally, but if enough people think it's a deal breaker then this proposal doesn't really have legs. Let me put up a forum post after I do some cleanup and we can have a larger discussion there.

jclapis: I added a clarification that this only applies to nodes with an RPL withdrawal address set. Nodes without one are not affected, and this won't change the behavior they already use today.

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,162 @@
...
+
+Realistically, most node operators won't end up using this feature. It's only really helpful for node operators that don't supply their own RPL when creating a minipool. If you don't fall into this camp, rest assured that *this proposal will have no impact on you*.
+
+Unfortunately, node operators that *do* use this new feature can **no longer trigger rewards claims** from the Merkle rewards system. More specifically they can distribute minipool rewards from the Beacon Chain, but can't claim the rewards produced every 28 days via from the Merkle rewards system. That's not a problem for RPL rewards, since the RPL withdrawal address gets those anyway, but the issue is this includes Smoothing Pool ETH rewards as well. Neither the node not the address that would get the Smoothing Pool ETH (the primary withdrawal address) can actually trigger a claim; *only* the RPL withdrawal address will be able to do it.
```
xrchz, line 19: "via from" redundant word

xrchz: "node not" - should be "nor"?

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,162 @@
...
+
+This can't be fixed for nodes that opt into the RPL withdrawal address set if they also use their node as their primary withdrawal address because of technical reasons. However, it *can* be fixed for nodes that opt into it and use an address *other* than their node as the primary withdrawal address. I anticipate that most nodes opting into the RPL withdrawal address feature will fall into this camp.
```
xrchz, line 21: "opt into the RPL withdrawal address set" is kinda confusing language. Suggestion: "opt into setting an RPL withdrawal address"

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,162 @@
...
+To be clear, the RPL withdrawal address is **not** the same setting as the primary withdrawal address. If a node has a "regular" withdrawal address set in Atlas, it *will not immediately be affected by this change*. Its current withdrawal address will become its primary withdrawal address and it can still claim rewards normally.
+
+The only nodes affected by this change are ones that explicitly set an **RPL withdrawal address** after Houston has been deployed, which is a different setting than the primary withdrawal address.
+
+
+## Specification
+
+The specification is an amendment to v9 specified in [candidate RPIP-52](https://github.com/rocket-pool/RPIPs/pull/174) by Patches. That proposal is **required** to be accepted prior to / at the same time as this one. 
+
+The formal specifications for rewards calculation, tree generation, and the corresponding rewards fileare included in [the assets folder](../assets/rpip-53/). More specifically:
```
xrchz, line 64: missing space "fileare"

RPIPs/RPIP-53.md
xrchz, line 21: Are these technical reasons ever explained? Is it the note at the bottom about avoiding an extra transaction? I don't think I agree with this decision if so.

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,162 @@
...
+
+As a note, the rationale for the **or** condition in the logic is purely designed to reduce unnecessary transactions. If nodes without a primary withdrawal address followed the same "breakout" rule that resulted in two sets of claimers, the RPL withdrawal address would have to claim for that node twice: once for itself (the RPL portion) and once for the ETH (the node portion). The node wouldn't be able to claim its own ETH portion since it still has an RPL withdrawal address set, and that's what the contracts key on when determining whether or not a node is eligible to claim - hence why this proposal exists in the first place. The easy way to mitigate this issue is just to set a primary withdrawal address to something other than the node; a practice that every node should be employing anyway if at all possible.
+
+
+## Reference Implementation
+
+I've implemented v10 into a Smart Node branch [here](https://github.com/rocket-pool/smartnode/tree/rewards_v10). For comparison, here are the differences between v8 and v10:
+
+- [For the vanilla (non-rolling) generator](https://www.diffchecker.com/tP55l6Iw/)
+- [For the rolling record generator](https://www.diffchecker.com/acLvVhpq/)
```
xrchz, line 156: I think these diffs might need to be inlined or made assets to avoid external dependencies.

RPIPs/RPIP-53.md
jclapis: There was a long discussion about this in #governance - basically the problem is if you don't include this particular condition, there will be two entries created in the tree: one at the node address with the node's ETH, and one at the RPL withdrawal address for the RPL. Since the former is only claimable by the RPL withdrawal address according to the Houston contracts, it means the RPL withdrawal address has to claim twice: once for their RPL, and once for the node's ETH. There's no direct incentive for them to do the latter since it costs them gas and they get nothing out of it, so we decided it was best to keep both sets of rewards in a single entry to ensure the RPL withdrawal address *does* have an incentive to claim.

I ended up taking that out of the spec for brevity but I can add it back in if you think it makes sense to do so.

RPIPs/RPIP-53.md
jclapis: Ok, replaced diffchecker with some vimdiffs in the assets.

assets/rpip-53/merkle-tree-spec.md
```diff
@@ -0,0 +1,164 @@
+# Specification for the Rewards Merkle Tree
+
+Rocket Pool's [Redstone](https://medium.com/rocket-pool/rocket-pool-the-merge-redstone-601d9efd6b4) update introduced a new rewards system based on a Merkle Tree distribution system.
+For each rewards interval, this tree will capture the RPL rewards (both for node operation and Oracle DAO membership) and the ETH rewards (from the Smoothing Pool, if relevant) for all of the nodes in the Rocket Pool network.
+
+In the [Houston](https://github.com/rocket-pool/docs.rocketpool.net/blob/houston/docs/guides/houston/whats-new.md) update, following [RPIP-53](https://github.com/rocket-pool/RPIPs/blob/rewards_v10/RPIPs/RPIP-53.md) the tree was modified slightly to account for the new [RPL withdrawal address](https://github.com/rocket-pool/RPIPs/blob/main/RPIPs/RPIP-31.md) feature.
```
xrchz, line 6: this kind of link seems at risk of going stale - can you do it with a commit hash instead?

and minor: the rpip links could be relative? absolute is also probably fine though

assets/rpip-53/merkle-tree-spec.md
```diff
@@ -0,0 +1,164 @@
...
+
+In total, the metadata will be an array of **116 bytes**.
+
+For example, say there were a claimer with the following values in its metadata:
+- Address: `0x8B0EF9f1932A2e44c3D27bE4C70C3BC07A6A27B3`
+- Network: `0`
+- Total RPL: 1.5 RPL (`1500000000000000000` in wei which is `14d1120d7b160000` in hexadecimal)
+- Total ETH: 2.5 ETH (`2500000000000000000` in wei which is `22b1c8c1227a0000` in hexadecimal)
+
+This tree node would have the following metadata byte array (in hexadecimal):
```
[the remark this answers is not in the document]
xrchz: Or maybe because you're trying to talk about the content of the tree node itself... you should still maybe say "leaf node"? internal tree nodes don't have this stuff..

assets/rpip-53/rewards-calculation-spec.md
```diff
@@ -0,0 +1,711 @@
+# Specification for Rewards Calculation
+
+This document serves as a formal specification for the way that the rewards intervals and the values within are calculated as part of the [Redstone](https://medium.com/rocket-pool/rocket-pool-the-merge-redstone-601d9efd6b4) rewards system.
+
+
+## Version
+
+This describes **v10** of the rewards calculation ruleset.
+
+This version implements [RPIP-53](https://github.com/rocket-pool/RPIPs/blob/rewards_v10/RPIPs/RPIP-53.md).
```
xrchz, line 10: will have to fix this link later. unless we make it relative here.

assets/rpip-53/rewards-calculation-spec.md
```diff
@@ -0,0 +1,711 @@
...
+
+### Changes since `v8`
+
+The following updates have been made from [v8](https://github.com/rocket-pool/RPIPs/blob/main/assets/rpip-51/rewards-calculation-spec.md) of the spec. Note that v9 was just a file spec change, not a ruleset change, and is thus not included in this comparison.
+
+
+#### Major Updates
+
+- The rewards tree no longer represents an individual Rocket Pool node in each leaf node; this concept has been generalized to a **claimer** which has eligible rewards and can claim them from the system. A claimer *may or may not be* a registered Rocket Pool node.
+- The process for determining claimers and their rewards has been introduced as part of rewards calculation, in [Determining Claimers and Claimer Rewards](#determining-claimers-and-claimer-rewards) section.
```
xrchz, line 21: add "the" between "in" and "[Determining"?

assets/rpip-53/rewards-file-spec.md
```diff
@@ -0,0 +1,103 @@
+# Specification for Rewards Tree File
```
xrchz, line 1: this was fixed (File -> Files) upstream - did you want to change it back, or just need to rebase?

jclapis: Yep, rebased with the latest asset from RPIP-52 - thanks!

assets/rpip-53/rewards-calculation-spec.md
jclapis: Made it relative here.

assets/rpip-53/merkle-tree-spec.md
jclapis: I ended up replacing it with a Medium article instead - I used that link in the v2 Beta 1 release so it makes more sense to use that here instead of the docs page.

xrchz (approved):
Looks good after several iterations of feedback and changes as an RPIP! Thanks to the authors (esp. jclapis) for their prompt and professional cooperation.

Minor comments that don't need changes: the abstract is kinda long, and there were a few cases of fixups in the specs that aren't strictly for this RPIP, but I think they're all worth adding on.

As for the RPIP itself I think there is potentially a major problem with the handling of RPL Withdrawal addresses that are also registered nodes, but we can leave that for the discussion period with the rest of the pDAO.

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,192 @@
...
+
+This RPIP proposes to achieve this by generalizing the rewards tree, changing *nodes* to *claimers*. The tree will now specify the amount of RPL and Smoothing Pool ETH available to individual claimers, where:
+- If a node does not have an RPL withdrawal address set, **or** if it has one but its primary withdrawal address is the same as the node address, their node address is the claimer address (same as today).
+- If a node has an RPL withdrawal address set and its primary withdrawal address is different from the node address, the node's rewards will be split in two:
+  - One claimer will be the node's primary withdrawal address, which will have the node's ETH portion of the rewards.
+  - The other will be the node's RPL withdrawal address, which will have the node's RPL portion of the rewards.
```
Valdorff, line 27: I'd probably keep this much simpler and point to Motivation for details.

> In the initial implementation there was an unforeseen issue where sometimes the parties receiving rewards would not be able to execute the claims to get them and would rely on another party (see (Motivation)[#motivation) for details). This RPIP proposes to mitigate the problem as much as possible by generalizing the rewards tree...

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,192 @@
...
+- Second, the node is still going to be listed in the rewards file, and thus the rewards for the corresponding interval are still going to show up as a "claimable interval" in the Smart Node despite the user not being able to claim from it, and not being able to actually *receive* any of the rewards from that interval. This can obviously be worked around in the Smart Node but it's exactly that: a workaround. The file itself is still going to have the node operator listed, so anyone (such as a third party) that wants to use it to view or claim rewards has to know about and implement this workaround as well.
+
+To make it clear, under the Houston contracts here is the set of rules for claiming rewards:
+
+
+### Vanilla Houston - Rewards for Interval X
+
+*Unless otherwise specified, withdrawal addresses are assessed at the time of the claim, not the time of Interval X.*
+
+| Primary Withdrawal Address | RPL Withdrawal Address Set? | Who Can Claim?   | Who Gets ETH? | Who Gets RPL? |
```
Valdorff, line 57: Can we word "Who Can Claim?" differently? Usually if an entity claims something, it's for themselves. Maybe "Who Can Initiate Claims?" or "Who Can Send Claim Transaction?"?

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,192 @@
...
+| -------------------------- | --------------------------- | ---------------- | ------------- | ------------- |
+| Same as Node               | No                          | Node             | Node          | Node          |
+| Different from Node        | No                          | Node, Primary WA | Primary WA    | Primary WA    |
+| Same as Node               | Yes                         | RPL WA           | Node          | RPL WA        |
+| Different from Node        | Yes                         | RPL WA           | Primary WA    | RPL WA        |
+
+The issue here is with row 3 and 4 (though we can only affect row 4).
+
+In this RPIP, I'm proposing a small change to the rewards file that ameliorates the problems when possible and provides some helpful benefits for people that would actually take advantage of the feature in the first place. Here's what it would look like under the new changes:
```
Valdorff, line 66: Minor: "This RPIP proposes a small..."

RPIPs/RPIP-53.md
```diff
@@ -0,0 +1,192 @@
...
+
+While this condition is necessary for Houston, it could be revised or even eliminated in Saturn if the Protocol DAO expresses interest in fixing it at the contract level.
+
+
+## Reference Implementation
+
+I've implemented v10 into a Smart Node branch [here](https://github.com/rocket-pool/smartnode/tree/rewards_v10). For comparison, here are the differences between v8 and v10:
```
Valdorff, line 183: Minor: v10 has been implemented into a Smart Node branch [here]...

RPIPs/RPIP-53.md
jclapis: I changed it to "Who Can Trigger Claims?" since that's how the process is referred to in the rest of the proposal.

RPIPs/RPIP-53.md
jclapis: Ok, shortened the abstract - hopefully it reads better now.

PR #182, opened 2024-05-02
Update RPIP-53 with Appendix

jclapis:
This adds an appendix of how various unintuitive scenarios will play out in accordance with v10 (and in some cases, with vanilla Houston for posterity). Also adds a bit of clarity surrounding the ordering of nodes when assessed in the claimer construction.


RPIPs/RPIP-53.md
```diff
@@ -183,5 +183,136 @@ v10 has been implemented into a Smart Node branch [here](https://github.com/rock
...
+
+```
+Entry A (A's ETH, A's RPL)
+
+Entry B (B's ETH, B's RPL)
+```
+
+- Eligible claimers for Entry A and Entry B are both determined at the time of claiming.
+
+Note that if this is the scenario at the time of claiming Entry A, then only B can claim Entry A. However, the rewards will go to A and C. B itself won't earn anything and it will have to pay for gas, so this situation only works trustlessly when the owner of B also owns C.
```
xrchz, line 210: Why does C get node A's rewards?

RPIPs/RPIP-53.md
```diff
@@ -183,5 +183,136 @@ v10 has been implemented into a Smart Node branch [here](https://github.com/rock
...
+```
+Entry A (A's ETH, A's RPL)
+
+Entry C (B's ETH)
+
+Entry D (B's RPL)
+```
+
+- Eligible claimers for Entry A are determined at the time of claiming.
+- Entry C will only be claimable by C for all time (unless C become a node operator prior to claiming, in which case the typical node operator rules for Houston apply).
```
xrchz, line 234: become -> becomes

RPIPs/RPIP-53.md
jclapis: Looks like that's a mistake, thanks for the catch. Updated to reflect that B will get the RPL here.

RPIPs/RPIP-53.md
```diff
@@ -183,5 +183,136 @@ v10 has been implemented into a Smart Node branch [here](https://github.com/rock
...
+Entry A (A's ETH, A's RPL)
+
+Entry B (B's ETH, B's RPL)
+```
+
+- Eligible claimers for Entry A and Entry B are both determined at the time of claiming.
+
+Note that if this is the scenario at the time of claiming Entry A, then only B can claim Entry A. Even though B has a primary withdrawal address set (to C), B will still receive the rewards upon claiming. The contract doesn't look at B's withdrawal addresses when it's claiming for A as A's RPL withdrawal address, because the contracts don't check whether or not B is a node during that process.
+This is the same with or without v10, but is included here for posterity. 
+
```
xrchz, line 212: "Entry B is straightforward following Houston rules: if the scenario is as described at the time of claiming, Entry B can be claimed by B or C and all funds will go to C."

May be nice to add this just to be explicit about what the normal Houston idea is (referenced several times below too).

jclapis: Ok, updated with a slight variant of that to clarify how both entries work under Houston rules.

xrchz (approved):

PR #185, opened 2024-05-03
Flip ordering on tokenomics page so it's oldest first.

LongForWisdom:


htimsk (approved):
Simple edit to flip the ordering from newest first to oldest first.

PR #194, opened 2024-05-24
Changed two guardrail ranges from exclusive of a bound to inclusive. …

Valdorff:
…This has no expected impact, and matches the actual implementation.


xrchz:
The `security.members.quorum` parameter does not exactly match the code in two ways: (a) the code calls it `members.quorum` (no security prefix) -- although maybe this is intended as it's part of the security contract?, (b) the code comments use < (though the code itself uses <=).

xrchz (approved):

htimsk (approved):
Good fix adding two "=" signs.

PR #196, opened 2024-05-26
Node distributor balance tracking

knoshua:


xrchz:
The title feels a bit off - "tracking" sounds a lot more passive than what this is. Maybe something like: Unbundling user funds in node distributors.

xrchz (approved):

nichechristie (commented):
I'm awesome

PR #197, opened 2024-05-28
Added vote information for RPIP-52.

LongForWisdom:


LongForWisdom:
Bleh, meant to separate these but committed to the wrong branch and CBA to revert. Still an update so w/e.

PR #199, opened 2024-05-28
Balance Submission Guardrail RPIP

knoshua:


RPIPs/rpip-balance-sub-guard.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Maximum rETH Delta Protocol Parameter
+- There SHALL be a Maximum rETH Delta protocol parameter that the pDAO can vote to change.
+- Maximum rETH Delta SHALL initially be set to 2 percent.
+- The pDAO SHALL NOT be able to set Maximum rETH Delta lower than 1 percent.
+- The Security Council SHALL NOT be able to change Maximum rETH Delta.
+
+### Network Balance Updates
+- Network Balance updates SHALL NOT be able to be sumbitted more frequently than specified by the existing protocol parameter Network Submit Balances Frequency.
+- Network Balance updates SHALL NOT be able to change the rETH exchange rate by more than Maximum rETH Delta.
```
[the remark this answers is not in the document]
knoshua: I'm going with:
> The oDAO SHALL NOT be able to change the rETH exchange rate by more than Maximum rETH Delta.

I think the "percent" is unnecessary here because Maximum rETH Delta is already defined as a percent value above.

RPIPs/rpip-balance-sub-guard.md
```diff
@@ -0,0 +1,52 @@
...
+## Specification
+
+### Maximum rETH Delta Protocol Parameter
+- There SHALL be a Maximum rETH Delta protocol parameter that the pDAO can vote to change.
+- Maximum rETH Delta SHALL initially be set to 2 percent.
+- The pDAO SHALL NOT be able to set Maximum rETH Delta lower than 1 percent.
+- The Security Council SHALL NOT be able to change Maximum rETH Delta.
+
+### Network Balance Updates
+- The oDAO SHALL NOT be able to be submit a Network Balance update if the time elapsed since the last Network Balance update is less than the Network Submit Balances Frequency.
```
Valdorff, line 30: I'm a little late to the party... but shouldn't this be about target block rather than submission time?

An attempt:
```
- The oDAO SHALL NOT be able to be submit a Network Balance update if the time elapsed between the target block and the last successful Network Balance update's target block is less than the Network Submit Balances Frequency.
```

knoshua: If we go to target block as suggested, this opens the door to:
- no updates happen for a a week
- within one block, 7 updates happen with target blocks Monday, Tuesday, Wednesday, ...

So I do think I prefer directly limiting how often submissions can happen.

Valdorff (approved):
Looks good and my concern has been resolved

PR #201, opened 2024-05-31
Formatted version of first tokenomics rework explainers. Mutex with bulleted version

LongForWisdom:
I can't have both versions on staging at once. I prefer this version. So staging link here: <https://longforwisdom.github.io/RPIPs/tokenomics-explainers/001-why-rework>

image

image

Opted to split the first high-level doc in half because all together it's fairly long for a high-level explainer, and there is a very clear point of separation. Given they're on the portal, I can and have made it really easy to navigate between them.


LongForWisdom:
Note that I have rewritten some sentences in here to make it work better not-as-bullets. Flow is almost entirely the same, but fair warning to review the actual text as it's not 1 to 1 from the source anymore.

(Realized I put the above comment on the wrong PR)

PR #205, opened 2024-06-04
Keeping RPIP-49 up to date.

LongForWisdom:


Valdorff:
Maybe we should swap to checkmarks? That way 1 and 3 can be shown as done and 2/4 as ongoing?

LongForWisdom:
> Maybe we should swap to checkmarks? That way 1 and 3 can be shown as done and 2/4 as ongoing?

Sure, lets find out if checkmarks work lol.

LongForWisdom:
They do not. Went with this:
image

Valdorff (approved):
ty!

PR #218, opened 2024-06-13
Catch up spring2024-tokenomics before merging it to main

orangesamus:


Nichebiche (commented):
Ethereum is everything good

PR #220, opened 2024-06-17
Minor fixes to rpip-33

Valdorff:
- Clarify namespacing for settings
- Remove "balance range" -- that setting has not been implemented
- Remove "security." prefix that's not actually used for a handful of settings


htimsk (approved):
Clarifying changes look good.

PR #221, opened 2024-06-17
[TOKENOMICS, DO NOT MERGE] - Draft6 Line Comment Feedback!

LongForWisdom:


RPIPs/RPIP-60.md
```diff
@@ -1,52 +0,0 @@
...
-  - The security council SHALL use this power if and only if:
-    - A vote specifically requested a fast upgrade
-    - The vote passed with a supermajority of `fast_upgrade_supermajority` or more
-    - The security council deems it acceptable from a security point of view
-- The security council SHALL have the power to veto contract upgrades (which are proposed by the oDAO)
-  - Veto powers SHALL NOT be used lightly and SHALL be reserved for cases of vote manipulation, malicious action (eg, oDAO votes in a proposal against the pDAO governance), or contract upgrades that would result in clear damage to the Rocket Pool project
-  - When this veto power is exercised, the security council MUST publish a Veto Explanation Document that describes why this step was taken; if possible, the report SHOULD also suggest potential similar-but-non-damaging votes that could be considered
-  - If this veto power is abused, the pDAO SHOULD consider replacing the security council
-- The initial settings SHALL be:
-  - `default_upgrade_delay`: 3 weeks
```
LongForWisdom, line 41: Does this need to exceed the unstaking delay imposed on RPL staked within nodes? Imo, there should be a way for RPL staking node operators to entirely bail from the protocol before the update takes effect. 

Given having a 6 week delay (2 weeks for visibility, 4 weeks for unstaking delay) seems really long, perhaps the ability to skip the unstaking delay by citing the ID of an upgrade currently queued in the delay system?

RPIPs/RPIP-60.md
```diff
@@ -1,52 +0,0 @@
...
-  - `fast_upgrade_delay`: 3 days
```
LongForWisdom, line 42: Probably too slow. I would rather see like 24 hours alongside a unanimous requirement.

RPIPs/RPIP-60.md
```diff
@@ -1,52 +0,0 @@
...
-## Specification
-
-### Protocol upgrades
-- All protocol upgrades SHALL have a `default_upgrade_delay` delay between when they are passed and when they are executed
-- The security council SHALL have a limited-use power to use a shorter delay of `fast_upgrade_delay`
-  - This power SHALL be usable if `fast_upgrade_seal_count` > 0 
-  - `fast_upgrade_seal_count` SHALL be decremented by one upon using this power
-  - The pDAO MAY include a change to `fast_upgrade_seal_count` in a voted upgrade*
-  - The pDAO MAY change `fast_upgrade_seal_count` via vote**
-  - The security council SHALL use this power if and only if:
```
LongForWisdom, line 32: Can we add something here that smart contract upgrade code:
* Has been public for at least x time? 
* Is verified and such on etherscan?

Maybe these can be taken as obvious assumptions? Not sure.

Thinking about how this actually works, if with requires a pDAO vote, those are like 2 weeks lead time, so maybe a requirement that the upgrade code is _complete and public_ during that pDAO vote? Like, its less good to have pDAO voting on something that's not been written yet.

RPIPs/RPIP-44.md
```diff
@@ -1,56 +0,0 @@
...
-- A megapool contract SHALL include an `exit` function
-  - This function SHALL accept one or more public keys for validators associated with this megapool
-  - This function SHALL either remove the specified validators from the megapool or revert
-      - If any of the validators are still active on the beacon chain a successful call of this
-        function MUST also exit those validators from the beacon chain
-  - This function MAY be freely called by the Node Operator (as implied by [RPIP-43](RPIP-43.md))
-  - This function MAY be called by any account under the condition: `deficit >= exit_deficit`
-      - It MUST NOT be possible for accounts other than the Node Operator to exit more validators than needed to reduce `deficit` below `exit_deficit`
-- The protocol SHALL use the `withdrawn`, `credit`, and `received` balances to pay off `debt` prior to taking action on exits
-- The protocol SHOULD use staked RPL to decrease `debt` by the corresponding amount prior to taking action on exits
```
LongForWisdom, line 53: Is it just me, or is there a lot of complexity here? You need to swap it to ETH at the point it decreases debt? Otherwise if you hold it in RPL protocol loses out if price drops?

Could someone abuse this somehow?

RPIPs/RPIP-44.md
```diff
@@ -1,56 +0,0 @@
...
-
```
LongForWisdom, line 54: Needs rationale and security considerations.

RPIPs/RPIP-44.md
```diff
@@ -1,56 +0,0 @@
...
-```
-
-The exit functionality for a megapool is specified as follows:
-- A megapool contract SHALL include an `exit` function
-  - This function SHALL accept one or more public keys for validators associated with this megapool
-  - This function SHALL either remove the specified validators from the megapool or revert
-      - If any of the validators are still active on the beacon chain a successful call of this
-        function MUST also exit those validators from the beacon chain
-  - This function MAY be freely called by the Node Operator (as implied by [RPIP-43](RPIP-43.md))
-  - This function MAY be called by any account under the condition: `deficit >= exit_deficit`
```
LongForWisdom, line 50: This needs to be incentivized in some way for it to be functional absent the team (who would presumably call it periodically.) That can be handled via a wrapper though.

RPIPs/RPIP-45.md
```diff
@@ -1,59 +0,0 @@
...
-
-## Abstract
-This proposal establishes a contract that holds ETH from protocol revenue and allows for RPL to be burned in exchange for that protocol ETH. This provides direct utility to the RPL token.
-
-
-## Specification
-- A share of the revenue from borrowed ETH SHALL be directed to a smart contract (the burn contract) for this burn (see [RPIP-46](RPIP-46.md))
-- Any user MAY call a function in the burn contract to get ETH from the burn contract in exchange for RPL
-- Any RPL exchanged this way SHALL burned using the RPL token's `burn` or `burnFrom` functions
-- The swap price SHALL be based on an on-chain oracle
```
LongForWisdom, line 22: Need some margin on here if you expect this to be called regularly.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-
-## Abstract
-Currently, the commission determines the payout of revenue split between rETH and each specific minipool. Other parties, such as RPL, gain value indirectly. This proposal allows for splitting revenue between four initial parties: reth (the main product), node operators (the decentralized operators actually staking), voters (a subset of operators that have vote power), and a surplus revenue mechanism.
-
-This proposal also includes a small set of items for potential future use:
-- Seals that the security council can use to increase the NO share -- this is being used to find a reasonable setting based on the actual market.
-- An allowlist of controllers that may make changes to the settings, which allows for potential automation in the future
-
-## Specification
-1. Inflation settings SHALL be modified to retain inflation to the DAOs and eliminate inflation to NOs
```
LongForWisdom, line 23: Separate this into a subheading and drop it to the bottom of the specs. It is unrelated to the UARS mechanism.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-   1. `rpl.inflation.interval.rate` SHALL be set to `1000040763630249500` (1.5% per year)
-   2. Node Operators (`rocketClaimNode`) allocation SHALL be set to 0%
-   3. pDAO (`rocketClaimDAO`) allocation SHALL be set to 95%
-   4. oDAO (`rocketClaimTrustedNode`) allocation SHALL be set to 5%
-2. There SHALL be the following defined shares with settings: `node_operator_commission_share`, `voter_share`, `surplus_share`
```
LongForWisdom, line 28: Are these pDAO settings? Security Council permission over them? etc. this one is fairly long, try to separate into headings, pDAO settings is maybe a good heading.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
----
-
-## Abstract
-Currently, the commission determines the payout of revenue split between rETH and each specific minipool. Other parties, such as RPL, gain value indirectly. This proposal allows for splitting revenue between four initial parties: reth (the main product), node operators (the decentralized operators actually staking), voters (a subset of operators that have vote power), and a surplus revenue mechanism.
-
-This proposal also includes a small set of items for potential future use:
-- Seals that the security council can use to increase the NO share -- this is being used to find a reasonable setting based on the actual market.
-- An allowlist of controllers that may make changes to the settings, which allows for potential automation in the future
-
-## Specification
```
LongForWisdom, line 22: Numbers versus bullets. I don't hate it, but would prefer consistency.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-   3. `surplus_share`: this share of revenue is used to distribute revenue beyond that used for protocol operation (such as the shares above)
-3. `reth_commission` SHALL be defined as the sum of all defined shares that have settings
-4. `reth_share` SHALL be defined as `100% - reth_commission`
-5. Distributions of revenue from borrowed ETH MUST respect the defined shares
-   1. If shares change between claims, distributions MUST make an effort to account for the different values. For example, a distribution could use a duration-weighted average share. Approximations MAY be used where they significantly reduce complexity and/or costs.
-   2. Legacy minipools are an exception and MAY continue to support earlier distribution methodologies 
-6. `node_operator_commission_share`, `surplus_share`, `increase_no_share_seal_increment`, `increase_no_share_seal_count`, and `allowlisted_controllers` MAY be updated by pDAO vote
```
LongForWisdom, line 37: Can move to new parameters section.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-7. `node_operator_commission_share` and `surplus_share`, MAY be updated by an address in the `allowlisted_controllers` array
```
LongForWisdom, line 38: To future proof, does it make sense to have a separate array per-setting?

LongForWisdom: eg if we want to automate voter_share without replacing the contract.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-   1. This functionality SHALL not be used without a separate pDAO vote to enable a controller and add it to the list
-8. The security council SHALL have a limited-use power to increase the `node_operator_commission_share` by `increase_no_share_seal_increment` and decrease the `surplus_share` by the same amount
```
LongForWisdom, line 40: Rather than have this system of seals, does it make more sense to just have the change on a timer? EG security council can move it up or down once every 14 days only, within the predefined min/max. Just seems a lot less complicated than seals and replenishment by pDAO.

LongForWisdom: If not, should be an entry in rationale section to explain maybe. Seems over-engineered as is.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-   1. This power SHALL be usable if `increase_no_share_seal_count` > 0
-   2. `increase_no_share_seal_count` SHALL be decremented by one upon using this power
-   3. This power SHOULD be used if the deposit pool is over half-full for the majority of a 2-week period at the current `node_operator_commission_share`
-   4. The pDAO MAY change `increase_no_share_seal_count` via vote
-9. The initial settings SHALL be:
```
LongForWisdom, line 45: Could be part of the 'new parameters' section.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-   1. `node_operator_commission_share`: 3.5%
-   2. `voter_share`: 5%
-   3. `surplus_share`: 5.5%
-   4. `increase_no_share_seal_increment`: 0.5%
-   5. `increase_no_share_seal_count`: 6
-   6. `allowlisted_controllers`: []
-10. Prior to the release of Saturn 1, a ranked-choice vote MUST be held to select a mechanism for surplus revenue share
```
LongForWisdom, line 52: Separate heading for this too, please.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-       1. If not, the revenue SHALL be held in reserve until the mechanism is implemented. In such a case, once the mechanism is implemented, the revenue SHALL be distributed in the same amount of time it took to build up, or faster.
-    4. The selected mechanism MUST be implemented in Saturn 2
-    5. When implemented, `surplus_share` SHOULD be renamed to something more specific
-
-## Specification taking effect with Saturn 2
-### Updating `voter_share`:
-The `voter_share` SHALL be adjustable, to target a share of voting eligible RPL (including 
-megapool staked RPL and legacy staked RPL) relative to the total supply of RPL minus RPL 
-owned by the protocol in its treasury or as a result of `surplus_share` buy backs. The voting
-eligible RPL target is a range between `vote_eligible_target_min` and `vote_eligible_target_max`
```
LongForWisdom, line 65: Defining a maximum target for actively voting RPL seems wrong from a governance POV. General metric of most governance systems is their level of turnout, with higher almost always being better. At the least, should it be higher than 65%? Like 90%? 95%?

RPIPs/RPIP-46.md
LongForWisdom: Alternatively, if we want to explicitly enable multiple changes in a larger burst, have seals replenish up to a limit once every 2-weeks.

LongForWisdom: Requiring pDAO to manually replenish seems like a tax we can avoid, they should only need to intervene if they disagree with the council's choices, rather than intervene to agree.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-      1. `voter_share` is decreased to `voter_share / (1+voter_share_relative_step)`
-      2. `surplus_share` is increased by the difference between the old and new `voter_share`
-   6. Because this involves _voters_ modifying `voter_share`, there is an acknowledged conflict of interest here. As a result, changing this method of "Updating `voter_share`" SHALL require a supermajority vote with at least 75% of the vote in support of any change.
-3. `voter_share_relative_step`, `vote_eligible_target_min`, and `vote_eligible_target_max` MAY be updated by pDAO vote; however, it SHALL require a supermajority vote with at least 75% of the vote in support of any change.
-4. The initial settings SHALL be:
-   1. `voter_share_relative_step`: 15%
-   2. `vote_eligible_target_min`: 55%
-   3. `vote_eligible_target_max`: 65%
-
-## Optional heuristics
```
LongForWisdom, line 84: I agree with Samus this should be removed. I think ideal is to move it to a supporting document that's linked and present in the repo.

RPIPs/RPIP-46.md
```diff
@@ -1,108 +0,0 @@
...
-- If the NO queue is continuously over 1000 deposits for 4 weeks and the trend is upwards, the pDAO should take action to decrease the supply of NOs.
-- If `increase_no_share_seal_count` reaches 1, the pDAO should (a) consider adding seals, (b) consider stepping up `no_share` themselves, and (c) consider decreasing `reth_share`
-- When there are large changes to the system (eg, Saturn 2 release), do note that some volatility is expected and should be considered when acting
-- If we are approaching the self-limits described in [RPIP-17](RPIP-17.md), the pDAO should act to limit one or both of rETH demand (via reducing RPL inflation spend on rETH demand and/or lower `reth_share`) or NO supply (via lower `no_share`). This would result in higher `surplus_share` (or lower RPL inflation).
-
-## Historic revenue share values
-| Date                         | Share Settings                                                                     |
-|------------------------------|------------------------------------------------------------------------------------|
-| 2024-03-08<br>(ratified TBD) | `node_operator_commission_share`: 3.5%, `voter_share`: 5%, `surplus_share`: 5.5%  |
-
```
LongForWisdom, line 106: Needs rationale and security considerations.

RPIPs/RPIP-47.md
```diff
@@ -1,46 +0,0 @@
...
-requires: 43
-tags: tokenomics-2024, tokenomics-content
----
-
-## Abstract
-Currently, Node Operators can choose to upgrade (or not) for their minipool delegate (the contract at the Ethereum withdrawal address that governs how funds are disbursed, among other things). This (a) means the protocol can only effect change in ways that are in the interest of individual NOs, and (b) makes it more challenging to design the protocol as it must remain backwards compatible with every past minipool delegate.
-
-This proposal suggests limiting that upgrade choice in the future. Users will be able to opt in to upgrade or use an "old" delegate for a defined period after a newer version. However, once the defined period has passed, they will no longer use the old delegate. To ensure timely availability of protocol funds, after the defined period, anyone may change the megapool's delegate to a minimal 'recovery delegate' that is only capable of exiting validators and distributing funds. 
-
-## Specification
```
LongForWisdom, line 20: Imo, you want a section to define the parameter. upgrade buffer should be pDAO settable, right?

RPIPs/RPIP-47.md
```diff
@@ -1,46 +0,0 @@
...
-### Megapool delegates
-- `Megapool delegate` contracts SHALL have an expiration block (ie, execution layer block)
-- When a new `megapool delegate` is released, the contract upgrade SHALL include:
-  - Setting the new `megapool delegate`'s expiration block to "no expiration"
-  - Setting the previous `megapool delegate`'s expiration block to occur `delegate_upgrade_buffer` after the upgrade
-- A node operator SHALL be able to upgrade their active `megapool delegate` to the latest `megapool delegate` at any time (including after the active `megapool delegate`'s expiration block)
-- After the active `megapool delegate`'s expiration block:
-  - The active `megapool delegate` SHALL NOT be usable by the node operator
```
LongForWisdom, line 28: Does not-usable need to be better defined? Does this block vote weight? Which functions? etc.

RPIPs/RPIP-47.md
```diff
@@ -1,46 +0,0 @@
...
-- The `megapool recovery delegate` for a megapool SHALL be usable by any address if the current block is greater than the active `megapool delegate`'s expiration block
-- The `megapool recovery delegate` SHALL have the following capabilities:
-  - Force exiting ALL validators under the megapool
-    - In the Saturn 1 release, this MAY be a non-functional stub
-  - Removal of any pending validators from the Rocket Pool queue
-  - Processing of voluntary exit proofs
-  - Collection of any penalties or deficits
-  - Final distribution of rewards and capital 
-- The `megapool recovery delegate` SHOULD be updated only rarely, and only to support changes to the Ethereum execution layer exit process, or major changes in the Rocket Pool protocol. Simultaneous upgrades to both the `megapool recovery delegate` and the latest `megapool delegate` SHOULD be avoided if possible.
-
```
LongForWisdom, line 44: Rationale, Security, etc.

LongForWisdom: Hard to follow the flows in this one, not sure how to resolve effectively. Maybe just needs a clear bullet that explains the impact on NOs? IE: Once you hit the expiry block, x, y or z will happen?

RPIPs/RPIP-59.md
```diff
@@ -1,90 +0,0 @@
...
-- There SHALL be an `express_queue`
-  - When adding a validator, users MAY place their deposit on the `express_queue` by spending one `express_queue_ticket`
-- When matching ETH from the deposit pool to queued deposits:
-  - First, ETH SHALL be matched to the oldest deposit in the `express_queue`; this is repeated `express_queue_rate` times
-  - Next, ETH SHALL be matched to the oldest deposit in the `standard_queue`
-  - This sequence SHALL be repeated indefinitely until both queues are empty
-    - If one queue is empty, those matches SHALL be skipped
-- Each node SHALL be provided `express_queue_tickets` equal to `express_queue_tickets_base_provision` (this includes newly created nodes)
-- Each node SHALL be provided additional `express_queue_tickets` equal to `(bonded ETH in legacy minipools)/4` (this will always be zero for newly created nodes)
-- The initial settings SHALL be:
```
LongForWisdom, line 36: Again, pDAO settings? Security Council has authority? etc.

RPIPs/RPIP-59.md
```diff
@@ -1,90 +0,0 @@
...
-- `deposit` SHALL place the Node Operator ETH in the deposit pool (where it can be used in validators as needed) and place the validator in a queue as described [above](#deposit-queue-specification)
-- The following values for the validator SHALL be stored on chain:
-    - the public key of the validator
-    - the BLS signature over the public key, the withdrawal credentials (the megapool address of the node operator), and the amount (1 ETH) as required by the deposit contract
-    - the deposit message root as required by the deposit contract
-- The transaction SHALL validate the provided values and revert if they do not match
-- `deposit` SHALL assign deposits as described below 
-
-#### Assigning ETH from the Deposit Pool
-- As ETH enters the deposit pool, it SHALL be assigned to valdiators from the queue by sending 32 ETH to the associated megapool contract
```
LongForWisdom, line 53: Might be this is obvious, and I just don't have enough understanding of the protocol. Why is this 32 ETH? Shouldn't it be 32 - bond provided by the operator?

RPIPs/RPIP-59.md
```diff
@@ -1,90 +0,0 @@
...
-- If `stake` is not called within `time_before_dissolve` after the ETH was assigned, the validator SHALL be dissolved, returning the user balance to the deposit pool
-
-#### Exiting Queue
-- Until ETH is assigned to a validator, it SHALL be possible to exit the queue and receive ETH `credit` for it
-
-#### Social Assignments
-- The deposit.assign.socialised.maximum setting SHALL be set to 0
-
-
-#### Initial Settings
```
LongForWisdom, line 69: Who can change, when, etc.

RPIPs/RPIP-59.md
```diff
@@ -1,90 +0,0 @@
...
-  - `scrub_period`: 12 hours
-  - `time_before_dissolve`: 2 weeks
-
-## Rationale
-
-### Deposit Queue
-- The express queue is meant to favor (a) small NOs and (b) existing NOs. The end goal in both cases is to support multiple values enshrined in [RPIP-23](RPIP-23.md) (the pDAO charter): decentralization, protocol safety, and the health of the Ethereum network.
-  - The `express_queue_tickets_base_provision` is enough to get started, and currently matches the length of `base_bond_array`
-  - The tickets from `(bonded ETH in legacy minipools)/4` are enough to fully migrate to 4-ETH deposits during Saturn 1 using the express queue OR to partly migrate to 1.5-ETH deposits after Saturn 2
-  - It's worth emphasizing that the tickets stick around -- ie, a node operator joining during a time when we don't have an NO queue (ie, when RP has an immediate need for NO supply) gets to keep their express queue benefit for a later time if they wish
```
LongForWisdom, line 80: Feels like what and not why. Maybe just reword. 'Tickets do not expire, because...'

RPIPs/RPIP-59.md
```diff
@@ -1,90 +0,0 @@
...
-  - When adding a validator, users MAY place their deposit on the `standard_queue`  
-- There SHALL be an `express_queue`
-  - When adding a validator, users MAY place their deposit on the `express_queue` by spending one `express_queue_ticket`
-- When matching ETH from the deposit pool to queued deposits:
-  - First, ETH SHALL be matched to the oldest deposit in the `express_queue`; this is repeated `express_queue_rate` times
-  - Next, ETH SHALL be matched to the oldest deposit in the `standard_queue`
-  - This sequence SHALL be repeated indefinitely until both queues are empty
-    - If one queue is empty, those matches SHALL be skipped
-- Each node SHALL be provided `express_queue_tickets` equal to `express_queue_tickets_base_provision` (this includes newly created nodes)
-- Each node SHALL be provided additional `express_queue_tickets` equal to `(bonded ETH in legacy minipools)/4` (this will always be zero for newly created nodes)
```
LongForWisdom, line 35: Is it worth saying something like: The pDAO can assign additional tickets to all NOs' to cover using this to manage future migrations?

RPIPs/RPIP-60.md
Valdorff: Right - this can't handle real emergencies. Some of the "pause" style powers may be able to. This would be more like "nobody can claim rewards -- it's super annoying, but the rewards are still all there to be claimed once resolved" or perhaps "fix the core issue so we can unpause"

RPIPs/RPIP-60.md
Valdorff: I'm not sure I'm tracking what "2 weeks for visibility" means. Also note that the vote itself is currently at least 2 weeks long (and only comes after required steps on the forum). I could see the case to change it to 30 days -- just a wee bit of buffer around the minimum. If you're in a weird edge case where you can't make those days work... maybe you start unstaking during the vote -- if it goes your way, you lose 1 month of RPL rewards -- hopefully there aren't many votes where the outcome is the difference between staying and leaving

RPIPs/RPIP-60.md
Valdorff: I'm not too concerned about pDAO vote's relationship to code. Our spec is generally higher level than our code by quite a lot, so it shouldn't impact the vote.

If it has not followed pDAO process, or the code cannot otherwise be checked for security -- those should not get proposed by the oDAO. If they are proposed without any verification, I would consider that malicious action from the oDAO and grounds for a security council veto.

RPIPs/RPIP-42.md
```diff
@@ -1,79 +0,0 @@
...
-- When a Node Operators creates a validator, with `i` validators in the megapool prior to adding: 
-  - If `i < base_bond_array.length`: the required `user_deposit` is the amount of additional ETH to bring the user's total bond up to `base_bond_array[i]`.
-  - If `i ≥ base_bond_array.length`:, the required `user_deposit` is `reduced_bond` per validator.
-- When a Node Operators removes a validator, with `i` validators in the megapool prior to removing:
-  - If `i > base_bond_array.length`: the Node Operator share before `debt` is `reduced_bond`.
-  - If `i ≤ base_bond_array.length` and `i > 1`: the Node Operator share before `debt` is the amount of ETH that would bring the user's total bond down to `base_bond_array[i-2]`.
-  - If `i==1`: the Node Operator share before `debt` is the amount of ETH that would bring the user's total bond down to 0 ETH.
-- Bulk validator creation/removal functions SHALL behave the same as multiple individual transactions.
-- If an NO has more total bonded ETH in their megapool than would be necessary based on the current settings (eg, `reduced_bond` is reduced), it SHALL be possible to reduce their bonded ETH and receive ETH `credit` for it
```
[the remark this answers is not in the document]
Valdorff: Yeah, it simply doesn't need to be supported. It would take force exiting to really support it, which seems wrong. 


❗ a note is worthwhile

RPIPs/RPIP-59.md
Valdorff: The security council has explicit control over specific settings. I don't think we need to go add "not controllable by the security council" to every setting. If "The initial pDAO settings SHALL be" will do the trick for ya, I think a one word adder doesn't bother me.

RPIPs/RPIP-59.md
Valdorff: The previous section placed the bond in the deposit pool (for use by the front of the queue), so the deposit pool is always sending 32.

RPIPs/RPIP-59.md
Valdorff: All pDAO settings are controlled by the pDAO by vote. Anything other than that needs to be explicitly specified.

RPIPs/RPIP-59.md
Valdorff: 👍 can do that. "We didn't want to make you use your good boi points to give us supply when we needed supply"

RPIPs/RPIP-59.md
Valdorff: I don't think that would be trivial to do, nor do I see a likely scenario for it (and if it's a migration I don't have a scenario for... hard to claim this is the best way).

RPIPs/RPIP-59.md
Valdorff, line 53: Minor: valdiators -> validators

LongForWisdom: How does this work if there are no rETH depositors, and several NOs trying to start validators? Do they all have credit, but if there are enough to hit 32 ETH, one can spin up a new validator? That would seem to break the whole 'you can exit the queue at any point and withdraw' thing defined elsewhere?

RPIPs/RPIP-59.md
LongForWisdom: Right, but do we even say these are pDAO settings in here?

RPIPs/RPIP-60.md
Valdorff: Some more context:

I recently learned that pDAO votes on-chain are a minimum of 3 weeks. `proposal.vote.delay.time` = 1 week between proposal creation and the start of voting, and then 2 weeks of voting. This is for settings and not being 

oDAO proposals right now have 1 week of delay before voting starts and 2 weeks of voting. They can be executed on as soon as quorum is reached (no need to wait to the end of the 2 weeks). In order to be legitimate (and thus not draw a veto from the security council), I'd expect it to be preceded by at least a 1-week snapshot (2-week recommended and I believe always used so far?) and a 1-week sentiment poll.

So it depends a little what we're counting:
- From sentiment poll going up, it's likely 4 weeks plus this setting
- From when the result is known, it is _just_ this setting

So if we want the latter to suffice, we'd need to do 30+ days.
30 days is quite a lot, especially if we get rid of the fast upgrade path (team has concerns about implementation complexity). I wouldn't really fight about it, but it's a real hit imo to be a whole month later to market on every upgrade.

RPIPs/RPIP-42.md
```diff
@@ -1,79 +0,0 @@
...
-- `credit` MUST be usable to create validators in a megapool
-- `credit` MUST be usable to mint rETH to the NO's primary withdrawal address 
-- The initial settings SHALL be:
```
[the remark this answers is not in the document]
Valdorff: RPIP-33 creates the security council and enumerates specific settings they control. I don't think we need to specify "this isn't a security council setting".

Maybe the correct move [which I'd suggest post-knoshua votes] is to strip it from those, and add a clarification line to RPIP-33 making it _abundantly_ clear that the security council only controls settings that are explicitly assigned to it and no others.

RPIPs/RPIP-42.md
```diff
@@ -1,79 +0,0 @@
...
-  - A moderate step would be to change `base_bond_array` to a curve that reduces MEV theft advantage in the current context (commission, MEV landscape...) at the cost of user complexity, eg `[4.2, 6.8. 9.2. 11.4. 13.5. 15.5. 17.4]`
-  - A larger step would be to pass EL rewards to NOs and charge them for the benefit. See eg: <https://github.com/Valdorff/rp-thoughts/tree/main/leb_safety#negative-commission-aka-assign-execution-layer-rewards-to-nos> or <https://dao.rocketpool.net/t/reimagining-large-block-theft/2146>
-  
-| 2.5% commission                                 | 4% commission                                |
-|-------------------------------------------------|----------------------------------------------|
-| ![img.png](../assets/rpip-42/theft_2.5pct.png)  | ![img.png](../assets/rpip-42/theft_4pct.png) |
-
-- Note that the bond curve has been chosen to ensure that minimum-size sock puppets are _not_ the most efficient for theft. They get the most _advantage_ from theft, but not the highest overall yield from a dishonest strategy.
```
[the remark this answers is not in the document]
Valdorff: Should say something like "minimum size sock puppets are most damaging to the protocol (ie, have the greatest drag), so it's important that they are not incentivized"

RPIPs/RPIP-42.md
```diff
@@ -1,79 +0,0 @@
...
-| MEV theft | theft size + 0.2 ETH | 2024-03-29 | 2024-03-29 |
-
-## Implementation thoughts
-- When showing legacy node status, there is not a trivial way to get the node index for a given address. That said, the other direction is trivial using `RocketNodeManager.getNodeAt`, so the work can get moved off-chain. Eg, one can iterate across all possible node indices and then pass in the node index that matches the node's address; the smart contract can confirm the match to demonstrate legacy node status.
-
-## Rationale
-- Bond sizes were originally ideated per [prior work](../assets/rpip-42/bond_curves.md).
-  - `base_bond_array` is chosen to "sufficiently" dissuade MEV theft as a strategy
-  - `reduced_bond` is chosen to "sufficiently" guard against slashing or abandonment risks
-- Follow-on work was done in [discord](https://discord.com/channels/405159462932971535/1228753782402318427/1228914436924772352)
```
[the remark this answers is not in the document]
Valdorff: I think this particular thing would fit under security considerations, actually

RPIPs/RPIP-42.md
```diff
@@ -1,79 +0,0 @@
...
-  - A larger step would be to pass EL rewards to NOs and charge them for the benefit. See eg: <https://github.com/Valdorff/rp-thoughts/tree/main/leb_safety#negative-commission-aka-assign-execution-layer-rewards-to-nos> or <https://dao.rocketpool.net/t/reimagining-large-block-theft/2146>
-  
-| 2.5% commission                                 | 4% commission                                |
-|-------------------------------------------------|----------------------------------------------|
-| ![img.png](../assets/rpip-42/theft_2.5pct.png)  | ![img.png](../assets/rpip-42/theft_4pct.png) |
-
-- Note that the bond curve has been chosen to ensure that minimum-size sock puppets are _not_ the most efficient for theft. They get the most _advantage_ from theft, but not the highest overall yield from a dishonest strategy.
-- Validators with `base_bond` deposits are prioritized to promote decentralization; new or smaller Node Operators can get up to `base_bond_array.length` validators launched ahead of larger Node Operators adding `reduced_bond` validators.
```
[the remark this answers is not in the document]
Valdorff: I think move to RPIP-59 entirely?

RPIPs/RPIP-43.md
```diff
@@ -1,172 +0,0 @@
...
-
-## Specification
-
-### Megapool Contracts
-
-This specification introduces a type of smart contract called a "megapool" that
-SHALL serve as the target of Beacon Chain withdrawal credentials for Rocket Pool
-validators deposited after this RPIP is implemented.
-
-- A Node Operator SHALL be able to deploy at most one megapool contract
```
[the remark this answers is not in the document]
Valdorff: Node Operator is just the controller of the of the node address, I don't think it implies a single human. I think we've used it consistently that way across many RPIPs (17, 32, 33, 4, ...)

RPIPs/RPIP-43.md
```diff
@@ -1,172 +0,0 @@
...
-       credentials to the megapool address
-- New and existing validators' withdrawal credentials MUST be checked to
-  correctly refer to the megapool address, with the validator status set to
-  Dissolved if the check fails
-     - The check MAY occur after the validator is added initially as Prestaked
-- A Node Operator SHALL be able to remove exited validators from their megapool
-
-### `debt` Variable
-- There SHALL be a `debt` variable that is initially set to 0
-- The oDAO SHALL be able to increase `debt` to apply penalties (see [RPIP-42](./RPIP-42.md/#penalizable-offenses)) by majority vote
```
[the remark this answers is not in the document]
Valdorff: Good idea as a "SHOULD"

RPIPs/RPIP-43.md
```diff
@@ -1,172 +0,0 @@
...
-- There SHALL be a reward distribution function in the megapool
-  - For this section, we define `borrowed_portion` as the megapool's `borrowed_eth / (bonded_eth + borrowed_eth)`
-  - When called, `reth_share * borrowed_portion` of rewards SHALL be sent to the rETH contract
-  - When called, `voter_share * borrowed_portion` of rewards SHALL be sent to a merkle rewards
-    distributor contract
-  - When called, `surplus_share * borrowed_portion` of rewards SHALL be sent to the appropriate
-    surplus disposition contract
-  - If `debt` exists when called, the remaining rewards SHALL first be used to pay off `debt`
-  - When called, any remaining rewards SHALL then be held in the megapool as unclaimed node operator funds 
-  - This function SHALL allow any user to call it
```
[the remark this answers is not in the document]
Valdorff: I like that vibe more, sure

RPIPs/RPIP-43.md
```diff
@@ -1,172 +0,0 @@
...
-- If the last unstake time is at least `unstaking_period` when additional
-  megapool staked RPL is having its state changed to Unstaking, the RPL already
-  in the Unstaking state SHOULD be withdrawn first
-
-### Vote Eligibility
-
-- All nodes are eligible to receive vote power
-- For nodes eligible to receive vote power, the sum of legacy staked RPL and
-  megapool staked RPL in the Staked state SHALL contribute to vote power up to
-  150% of the value of total bonded ETH (from both megapools and legacy
```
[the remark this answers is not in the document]
Valdorff: For the transition period, I think it's important that it be able to vote. I don't see much downside in leaving vote power intact long term either -- they have the needed mix of ETH and RPL to be aligned.

RPIPs/RPIP-44.md
```diff
@@ -1,56 +0,0 @@
...
-
-There is at least one additional use case not addressed in this RPIP, which is to exit badly-performing (eg, abandoned) validators. It is likely the pDAO will wish to supersede this RPIP to include that functionality. However, as it requires modeling and is not a critical component of the tokenomics rework, it is not being addressed at this time. For similar reasons, this RPIP does not include a keeper-based design to reward those that trigger forced exits. Please see a [historical version of RPIP-44](https://github.com/rocket-pool/RPIPs/blob/09d445accaa77f355acae1e943910ad0229a1d2e/RPIPs/RPIP-44.md) for initial but incomplete ideas to address abandonment and a keeper network.
-
-## Specification
-
-This specification extends the specification of megapools in
-[RPIP-43](RPIP-43.md) with a facility for removal of validators from the
-megapool to be done by parties other than the Node Operator.
-
-Funds (ETH) associated with a megapool SHALL be accounted in (at least) the following categories:
```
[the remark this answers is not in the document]
Valdorff: Hmmm... I tend to agree. RPIP-43 defines `rewards` as _not_ including principal and defines the no's share of `rewards`. It also says "capital funds" must be tracked. I think we might have a little disagreement between the two and need to make them mesh better. CC @knoshua

RPIPs/RPIP-44.md
Valdorff: There is definitely complexity. We do have an auction mechanism, technically, so that might be doable. Importantly this is a SHOULD, so it can be left off for schedule etc.

RPIPs/RPIP-44.md
Valdorff: Security considerations exist now; still needs rationale

RPIPs/RPIP-45.md
Valdorff: The premise was that the on-chain oracle, being on-chain, would need to use "time" averaging of some kind. So any time that there's a move down, this price will be a little outdated and thus provide margin.

RPIPs/RPIP-46.md
Valdorff: They are settings. pDAO and oDAO can both set settings. Security council only when explicitly allowed to control a specific setting.

RPIPs/RPIP-46.md
Valdorff: Agree. I like numbers in general cuz it makes it possible to refer to a line

RPIPs/RPIP-46.md
Valdorff: Not sure I understand. 1, 2, 6, 7, 9 are all directly setting-related and 3, 4 are definitions based on mathing settings. Are you looking to break out 5, 8, 10 into a separate section? 🤷 I guess that's fine; I don't have an opinion.

RPIPs/RPIP-46.md
Valdorff: I don't think that's necessary? The allowlisted controllers should be specific contracts with limited powers. Ie, we could vote to add a contract that only touches voter_share to the allowlist.

RPIPs/RPIP-46.md
Valdorff: That's pretty nice, yeah. There's also a setting-based variant without a seal power that was brought up in discord.

RPIPs/RPIP-46.md
Valdorff: As mentioned above -- the vast majority would be in a settings section. In fact, given that you've asked for 10 to be separated, I think only 5 and 8 are _not_ about settings.

RPIPs/RPIP-46.md
Valdorff: no strong opinion

RPIPs/RPIP-46.md
Valdorff: We won't have an easy metric for turnout, let alone "informed turnout". The 65% came from a tiny straw poll. I think this is actually a good candidate for the pDAO to select itself -- a sentiment poll could work, or even a full vote. This number doesn't dramatically impact the rest. There is a direct tradeoff with RPL having a shot at getting exogenous value at some point (eg, being usable as collateral for a loan).

Personally, 95% feels too high. 65-90% all seem reasonable to me, which I realize is quite a range.

RPIPs/RPIP-46.md
Valdorff: If we move it to somewhere less visible, it is gone for most intents and purposes. Eg, we have explicit documentation about our assumptions when choosing to go with 2.4ETH required for LEB8s. Our assumptions were violated and we took no action (and frankly didn't even explicitly realize it for many months).

In my eyes, it is very important to the idea of UARS that we _actively_ listen and react to the market, and exemplifying that is the point of this section.

RPIPs/RPIP-46.md
Valdorff: Still missing rationale, got security considerations

RPIPs/RPIP-47.md
Valdorff: To me, it already says that by calling it a setting.

RPIPs/RPIP-47.md
Valdorff: I think it's well defined as is? All functions on the megapool delegate that require the NO will be blocked. Actually... one thing that comes to mind is that we should block withdrawal address interactions too.

Voting doesn't use the megapool delegate, so it would still function. As long as money is properly stuck, I think the incentive is in place to fix it.

RPIPs/RPIP-47.md
Valdorff: Hmmm... I think that's there? Before expiration block: megapool delegate can be used, recovery delegate cannot. After: megapool delegate cannot be used, recovery delegate can be used.

RPIPs/RPIP-59.md
Valdorff: That's how it works, and you exit for credit, not ETH. There is a provision to trade credit for rETH (again, not ETH directly)

RPIPs/RPIP-59.md
Valdorff: There are no other setting types. There is only one thing we call settings. They can be changed by the pDAO or oDAO. They can be changed by the Security Council only if that right is explicitly given for that setting.

RPIPs/RPIP-42.md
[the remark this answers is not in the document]
LongForWisdom: Can do, yes.

RPIPs/RPIP-46.md
LongForWisdom: I've forgotten the exact scenario I was thinking here when I made the comment, but yeah, your point seems reasonable. Withdrawn.

RPIPs/RPIP-46.md
LongForWisdom: Yeah. That's fair. In practice might just end up worse.

RPIPs/RPIP-46.md
LongForWisdom: If the maximum changes while automation system is active, can that break things? You could have pDAO update it manually periodically. Otherwise yeah, makes sense to have them pick once at the start.

RPIPs/RPIP-46.md
LongForWisdom: If this section is that important. Should it become actual specification with SHOULD statements instead? eg:
"If the NO queue is continuously over 500 deposits for 2 weeks and the trend is upwards, the pDAO SHOULD act to either increase rETH demand or decrease NO supply via one or more of the following actions..."

RPIPs/RPIP-47.md
LongForWisdom: Should it be very explicit? "Functions in the active megapool delegate SHALL revert when not called by a protocol contract"?

RPIPs/RPIP-47.md
LongForWisdom: I feel like I have a better handle on what it is now. Chalk it up to needing sleep and being near the end of ~30 comments.

RPIPs/RPIP-59.md
LongForWisdom: Ah, okay. So they can exit the queue early, but they end up with rETH rather than ETH?

RPIPs/RPIP-59.md
LongForWisdom: There are oDAO settings and pDAO settings. Separately from the security council point here, I don't think we've said anywhere that these are pDAO settings. Like, the word 'pDAO' does not appear here, does it?

RPIPs/RPIP-60.md
LongForWisdom: Yeah. I don't disagree. It's long. I'm not sure how to reconcile that really. 

Realistically, you _should_ count just from when the result is known. Because exit prior to that is unlikely (people will hope the vote will go their way, and not act.) This is why I suggested skipping the withdrawal period by quoting the ID of a queued upgrade. We could set this to 48 hours then.

Valdorff: I don't see the point of that. If you can skip the withdrawal period, it removes much of the benefit of the withdrawal period (smoothing volatility, preventing runs, etc). It might only be a couple times a year, but they're important times.

RPIPs/RPIP-42.md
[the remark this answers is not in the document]
Valdorff: Lol -- it's actually already there! Looks like it got cloned during split.

RPIPs/RPIP-44.md
Valdorff: Changing to MAY

RPIPs/RPIP-46.md
Valdorff: Put in the setting-based variant. Not married to it per se.

RPIPs/RPIP-46.md
Valdorff: I think it would be safe, even if automation is on.

RPIPs/RPIP-46.md
Valdorff: They are meant to be truly optional (ie, we may learn better when we wish to act and how). OPTIONAL is an RFC2119 imperative similar to MAY, which is what I was trying to aim for her... I would love some help better communicating.

RPIPs/RPIP-47.md
Valdorff: I'm worried that might be too restrictive. Will try something like what I wrote above

RPIPs/RPIP-59.md
Valdorff: Crap, you're right. We do have a couple small sets of oDAO settings that the pDAO cannot change. The oDAO can change the pDAO settings (indirect or not, they can do it). I agree we should be saying "pDAO setting" in a thousand places. I will not be fixing that for the PR I'm working on right now.

RPIPs/RPIP-46.md
orangesamus: Imo the higher you go for "max", the less necessary/helpful the spec is (might as well not have an upper spec). The main concern is around "min" to ensure we have enough incentive for a sufficient amount of RPL staked. If we are using this spec that means there is a surplus share, and the general idea would be "pay enough to secure governance, and let the rest go to surplus share". The assumption is something along the lines of "we decided the mechanism used in surplus share provides the most value to RPL, so we want to give as much as possible to surplus share without sacrificing governance security (sufficient voter share)".

If 65% staked RPL is "enough" for governance, and the market staked 70%, then that means we are "over paying" voter share when it could be going to surplus share, so it should rebalance accordingly.

Also note that "staked RPL" doesn't equate to "level of turnout", you can stake RPL just for the yield but not be an active governance participant (not really what we are going for with "voter share"), so just because we end up with more staked RPL doesn't mean we are getting higher levels of voter turnout.

And finally this could all be useless if surplus revenue decision lands on "voter share" and we don't have a "surplus share" or automated targets here (and we may trend toward ~100% staked RPL)

RPIPs/RPIP-46.md
orangesamus: If we do keep this section in the actual RPIP then I think the disclaimers already present are good enough:
> "This section reflects some of the thinking at the time this RPIP was drafted. These ideas are explicitly not binding/enforceable, and they may freely change over time/context."... 

This can be a section we revisit before Saturn goes live and discuss further among nerds. If someone thoughtfully disagrees we should consider if there are better abstract guidelines to replace here, but since they are non-binding/enforceable it shouldn't block any work/time to market

RPIPs/RPIP-59.md
orangesamus: I agree generally settings should by default be controlled by pDAO and only need an explicit mention if that isn't the case (eg, controllable by security council"...)

RPIPs/RPIP-59.md
orangesamus: So we do need to explicitly state that? Why not just explicitly state when it's a oDAO setting and pDAO setting can be the default assumption...

RPIPs/RPIP-59.md
orangesamus: valdorff's pools can be called valdiators

RPIPs/RPIP-43.md
```diff
@@ -1,172 +0,0 @@
...
-This specification introduces a type of smart contract called a "megapool" that
-SHALL serve as the target of Beacon Chain withdrawal credentials for Rocket Pool
-validators deposited after this RPIP is implemented.
-
-- A Node Operator SHALL be able to deploy at most one megapool contract
-    - This MAY be combined with the deposit of their first validator
-- Megapool contracts MUST be upgradable using a proxy pattern; a megapool's
-  proxy implementation contract SHALL be called the megapool's "delegate"
-- Legacy minipool deposits (ie, deposits of validators with withdrawal
-  credentials that are not a megapool) SHALL be disabled
```
orangesamus, line 31: Should we use stronger language here and link to the definition anytime it is used in other RPIPs? Maybe change:
> - Legacy minipool deposits (ie, deposits of validators with withdrawal credentials that are not a megapool) SHALL be disabled

to something like:
> - We use the term "legacy minipool" to refer to a Rocket Pool validator with withdrawal credentials that are not a megapool
> - Legacy minipool deposits SHALL be disabled

"Legacy minipool" is also used in RPIP42, RPIP46, and RPIP59:
image

LongForWisdom:
@Valdorff I think you can just close this now. I will replace it once we're at a final review stage.

RPIPs/RPIP-43.md
LongForWisdom: Probably makes sense to unify this. IMO define legacy minipool once in RPIP-43, most of the others have 43 as a dependency anyway, so can be assumed to be that definition.

LongForWisdom:
Closing as said above. If you land here looking for a place to comment on the tokenomics proposals, feel free to DM or post in the  discord thread and we can direct you somewhere up-to-date.

PR #226, opened 2024-06-20
[TOKENOMICS, FOR REVIEW] - Added motivation sections to existing RPIPs

LongForWisdom:


RPIPs/RPIP-42.md
```diff
@@ -25,6 +25,14 @@ This proposal also explicitly tries to benefit the smallest NOs in a few ways, i
 - We give precedence to small nodes staking some number of initial validators in the Node Operator queue (see [RPIP-59](RPIP-59))
 - There is a small but tangible financial benefit for large stakers that stake as few large nodes instead of many small nodes -- this (alongside our vote power, which scales with the square root of vote-eligible RPL) helps preserve the strong governance voice of small NOs
 
+## Motivation
+
+The primary motivation for the introduction of smaller ETH bonds in this form is twofold. First, lower ETH bonds allow for the same amount of bonded ETH to support a larger volume of rETH, this helps unlock growth for the protocol, as it has been primarily limited by lack of Node Operators in the months preceding this proposal. Second, lower ETH bonds increase the profitability for Node Operators, allowing the protocol to grow and compete favorably with alternative liquid staking providers.
```
Valdorff, line 30: I'd tweak this a bit. Since we have UARs, it need not be NOs that get the additional profit. Rather, the overall system is more profitable, and then we steer it where needed (which could be NOs).
> Second, lower ETH bonds increase the capital efficiency of the protocol. The improved return per total investment allows the protocol to better compete with alternative liquid staking providers, eg, by paying node operators more.

RPIPs/RPIP-43.md
```diff
@@ -15,6 +15,14 @@ tags: tokenomics-2024, tokenomics-content
 ## Abstract
 This proposal drastically reduces the gas to add validators and distribute rewards for them. It does so by creating a single contract that can be used as the Ethereum withdrawal address for any number of validators. In addition, the contract will facilitate the application of node-level penalties, as opposed to the current state where penalties are assigned and processed at the minipool level. Note that this proposal also enables ETH-only Node operation.
 
+## Motivation
+
+Megapools are motivated by several desires; the desire to lower ongoing costs, to lower the barrier to entry, and to support improved capital efficiency. Both ongoing and upfront costs are reduced by Megapools due to the efficiency savings achieved by using a single smart contract over multiple minipool contracts. Megapools support improved capital efficiency indirectly by allowing for a shared stake that can be seized if _any_ of an operator's validators misbehave, whereas, in the current system of disconnected stake-per-minipool, the stake can only be seized from the offending minipool.
```
Valdorff, line 20: > Megapools allow lower initial and ongoing costs for node operators. Upfront costs are reduced by Megapools by using a single smart contract instead of multiple minipool contracts that need to be deployed separately. Ongoing costs are reduced by allowing consensus layer rewards to be distributed for many validators with a single transaction. These improvements are particularly important given the smaller bond sizes (and thus more validators per investment) proposed in [RPIP-42](RPIP-42).

It's possible to penalize across minipools as is (needs more accounting, but not necessarily a shared contract. The only capital efficiency improvements are via cost reduction, so focused more on that.

RPIPs/RPIP-47.md
```diff
@@ -17,6 +17,14 @@ Currently, Node Operators can choose to upgrade (or not) for their minipool dele
...
 
+## Motivation
+
+Supporting many iterations of the same basic functionality as the protocol evolves makes protocol development slower, more risky, and more expensive, with no real limit on the degree to which this debt can accumulate. The ability to force upgrades allows the protocol to avoid this tax on efficiency and remain responsive to future development needs.
+
+Additionally, the pDAO must have the ability to effect change in ways that benefit the protocol as a whole, but that may not benefit every individual node operator in isolation. What benefits individuals in the short term can lead to ruin in the long term for both that individual and the collective. The pDAO is unable to protect against this outcome without the ability to force upgrades or exits.
```
Valdorff, line 24: > without the ability to force upgrades or exits.

I'd probably just leave it at "upgrades" since that's the part relevant to this RPIP

RPIPs/RPIP-60.md
```diff
@@ -20,6 +20,14 @@ The security council may trigger a faster upgrade with a shorter delay, but ther
 
 Additionally the security council may veto contract upgrades. The veto power should not be used lightly, and misuse may result in replacement of the security council by the pDAO.
 
+## Motivation
+
+This RPIP is motivated by the desire to protect participants in the Rocket Pool as best as possible from unfavorable upgrades. Participants should have time to exit the protocol if it changes in a way they do not agree with, _prior_ to those changes coming into effect.
```
Valdorff, line 25: "participants in the Rocket Pool" -> "Node Operators"?

LongForWisdom: Thanks, typo. Holdover from when it was 'the protocol'

RPIPs/RPIP-42.md
LongForWisdom: Updated.

RPIPs/RPIP-42.md
```diff
@@ -25,6 +25,14 @@ This proposal also explicitly tries to benefit the smallest NOs in a few ways, i
 - We give precedence to small nodes staking some number of initial validators in the Node Operator queue (see [RPIP-59](RPIP-59))
 - There is a small but tangible financial benefit for large stakers that stake as few large nodes instead of many small nodes -- this (alongside our vote power, which scales with the square root of vote-eligible RPL) helps preserve the strong governance voice of small NOs
 
+## Motivation
+
+The primary motivation for the introduction of smaller ETH bonds in this form is twofold. First, lower ETH bonds allow for the same amount of bonded ETH to support a larger volume of rETH, this helps unlock growth for the protocol, as it has been primarily limited by lack of Node Operators in the months preceding this proposal. Second, lower ETH bonds increase the profitability of Rocket Pool as a protocol, allowing it to grow and compete favorably with alternative liquid staking providers.
+
+The innovation of a bond curve as opposed to a flat smaller allowable ETH bond is motivated by the need to maintain a base-level stake to mitigate MEV theft, slashing penalties, and abandonment; while meeting the core goal of improved capital efficiency.
```
orangesamus, line 32:
```suggestion
The innovation of a bond curve is motivated by the need to maintain enough stake to mitigate MEV theft, slashing penalties, and abandonment; while meeting the core goal of improved capital efficiency.
```

orangesamus: If the reader already knows what a bond curve is, then the "as opposed to" portion is unnecessary. If a reader doesn't know what a bond curve is, I don't think they will understand the concept just from "flat smaller allowable"...

If we wanted to provide a higher level explanation of what a bond curve is we could do that, but the explainer material already covers that, and then a more detailed description is found in the rest of the RPIP. So I think we can just omit as suggested

LongForWisdom: Sure, makes sense. It was a bit of an awkward phrasing before as well.

PR #239, opened 2024-07-08
[Don't merge!] MAYs that are ok

knoshua:
I highlighted MAYs that seem to make sense by appending ! for easier review.

One case I really wasn't sure what is meant, so I turned it into "MAY?"

### DIFF
--- a/RPIPs/RPIP-43.md
+++ b/RPIPs/RPIP-43.md
@@ -32,13 +32,13 @@ SHALL serve as the target of Beacon Chain withdrawal credentials for Rocket Pool
 validators deposited after this RPIP is implemented.
 
 - A Node Operator SHALL be able to deploy at most one megapool contract
-    - This MAY be combined with the deposit of their first validator
+    - This MAY! be combined with the deposit of their first validator
 - Megapool contracts MUST be upgradable using a proxy pattern; a megapool's
   proxy implementation contract SHALL be called the megapool's "delegate"
 - Legacy minipool deposits (ie, deposits of validators with withdrawal
   credentials that are not a megapool) SHALL be disabled
 - Solo migration into legacy minipools (ie, promotion of vacant minipools)
-  SHALL be disabled; this MAY be achieved by disabling creation of vacant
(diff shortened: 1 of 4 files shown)


RPIPs/RPIP-43.md
```diff
@@ -57,13 +57,13 @@ megapools SHALL include at least the following:
...
 - A Node Operator SHALL be able to add new validators to their megapool
-- A Node Operator MAY be able to migrate existing validators to their megapool
+- A Node Operator MAY! be able to migrate existing validators to their megapool
      - Note that this would involve updating the validators' BLS withdrawal
        credentials to the megapool address
 - New and existing validators' withdrawal credentials MUST be checked to
   correctly refer to the megapool address, with the validator status set to
   Dissolved if the check fails
-     - The check MAY occur after the validator is added initially as Prestaked
+     - The check MAY? occur after the validator is added initially as Prestaked
```
LongForWisdom, line 66: I think this is fine? I'd argue the whole clause is kind of unnecessary, but perhaps I'm missing something.

RPIPs/RPIP-43.md
```diff
@@ -88,7 +88,7 @@ into shares is defined in [RPIP-46](RPIP-46.md).
   The megapool MUST assume that any other funds present are rewards.
 - A Node Operator SHALL be able to distribute rewards from all validators in their
   megapool at once.
-- Rewards distribution MAY be temporarily blocked while validators are exiting or
+- Rewards distribution MAY! be temporarily blocked while validators are exiting or
```
LongForWisdom, line 91: This should maybe be clarified. Attempt:

"Rewards distribution MAY (fail gracefully / revert) while validators are exiting or pending removal."

RPIPs/RPIP-43.md
knoshua: Yeah, I'm confused why this clause is in there as well. Seems like the check hast to occur after but that's not really something we need to spec, because doing it before just isn't possible?

PR #240, opened 2024-07-09
Lfwfeedback scope etc

Valdorff:


RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
-   5. `increase_no_share_seal_count`: 6
-   6. `allowlisted_controllers`: []
-10. Prior to the release of Saturn 1, a ranked-choice vote MUST be held to select a mechanism for surplus revenue share
-    1. The choices MAY include [RPIP-45: RPL Burn](RPIP-45.md), [RPIP-50: RPL LP](RPIP-50.md), or using higher `voter_share`
-    2. If higher `voter_share` is selected, the entirety of the ["Specification taking effect with Saturn 2"](#specification-taking-effect-with-saturn-2) section below SHALL be deleted
-    3. The selected mechanism MAY be implemented in Saturn 1
-       1. If not, the revenue SHALL be held in reserve until the mechanism is implemented. In such a case, once the mechanism is implemented, the revenue SHALL be distributed in the same amount of time it took to build up, or faster.
-    4. The selected mechanism MUST be implemented in Saturn 2
-    5. When implemented, `surplus_share` SHOULD be renamed to something more specific
+6. The `node_operator_commission_share_council_adder` setting SHALL only be set to values within 0% and `max_node_operator_commission_share_council_adder` (inclusive)
```
sckuzzle, line 43: SHALL only have valid values of [...]?
As written this could be interpreted as the security council should do one thing without there being SC limits.

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
-4. `reth_share` SHALL be defined as `100% - reth_commission`
-5. Distributions of revenue from borrowed ETH MUST respect the defined shares
+2. `reth_commission` SHALL be defined as the sum of all defined shares that have settings
+3. `reth_share` SHALL be defined as `100% - reth_commission`
+4. Distributions of revenue from borrowed ETH MUST respect the defined shares
    1. If shares change between claims, distributions MUST make an effort to account for the different values. For example, a distribution could use a duration-weighted average share. Approximations MAY be used where they significantly reduce complexity and/or costs.
    2. Legacy minipools are an exception and SHALL continue to support earlier distribution methodologies 
-6. `node_operator_commission_share`, `surplus_share`, `increase_no_share_seal_increment`, `increase_no_share_seal_count`, and `allowlisted_controllers` SHALL be updateable by pDAO vote
-7. `node_operator_commission_share` and `surplus_share`, SHALL be updateable by an address in the `allowlisted_controllers` array
+5. `node_operator_commission_share` and `surplus_share`, SHALL be updateable by an address in the `allowlisted_controllers` array
```
sckuzzle, line 41: Should probably include `node_operator_commission_share_council_adder` in here as well.  This would allow the pDAO to increment NO commission and decrease the adder simultaneously, allowing council more room without changing the total NO commission.

RPIPs/RPIP-42.md
```diff
@@ -72,6 +72,11 @@ This portion of the RPIP SHALL be considered Living. It may be updated by a DAO
 - When showing legacy node status, there is not a trivial way to get the node index for a given address. That said, the other direction is trivial using `RocketNodeManager.getNodeAt`, so the work can get moved off-chain. Eg, one can iterate across all possible node indices and then pass in the node index that matches the node's address; the smart contract can confirm the match to demonstrate legacy node status.
 
 ## Rationale
+The bond curve is an attempt to get close to maximizing capital efficiency while maintaining safety. It balances capital efficiency with slashing, leakage, and MEV theft risks as described in [Security considerations](#security-considerations). 
+
+"Bond reduction" for credit is explicitly supported so that NOs that deposited under higher requirements are not permanently disadvantaged. At the same time, there is no equivalent support for "bond enlargement". This is because those validators are already running and it would need a heavy-handed approach, such as force exiting, to enforce bond enlargement. Since the bond curve is based on total bonded ETH, adding more validators would require the NO to join the active bond curve in such a case.
```
orangesamus, line 77: > adding more validators would require the NO to join the active bond curve in such a case.
Wording is a bit awkward. Do you mean if we decide to increase bond size then NO's now have to meet the additional requirement? Could delete that sentence altogether. Or rephrase:
> Since the bond curve is based on total bonded ETH, NO's would be required to add additional ETH bond to maintain their existing validators or add more validators

RPIPs/RPIP-46.md
```diff
@@ -15,9 +15,9 @@ tags: tokenomics-2024, tokenomics-content
 ## Abstract
 Currently, the commission determines the payout of revenue split between rETH and each specific minipool. Other parties, such as RPL, gain value indirectly. This proposal allows for splitting revenue between four initial parties: rETH (the main product), node operators (the decentralized operators actually staking), voters (a subset of operators that have vote power), and a surplus revenue mechanism.
 
-This proposal also includes a small set of items for potential future use:
-- Seals that the security council can use to increase the NO share -- this is being used to find a reasonable setting based on the actual market.
-- An allowlist of controllers that may make changes to the settings, which allows for potential automation in the future
+This proposal also includes:
+1. A security council-controlled settings that can mildly increase the NO share -- this is being used to find a reasonable setting based on the actual market more quickly than the pDAO is able to act
```
orangesamus, line 19:
```suggestion
1. A setting controlled by the security council that can mildly increase the NO share -- this can be used to react to market conditions more quickly than the pDAO is able to act
```

RPIPs/RPIP-46.md
orangesamus, line 43:
```suggestion
6. The `node_operator_commission_share_council_adder` setting SHALL only allow values within 0% and `max_node_operator_commission_share_council_adder` (inclusive)
```

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
-    5. When implemented, `surplus_share` SHOULD be renamed to something more specific
+6. The `node_operator_commission_share_council_adder` setting SHALL only be set to values within 0% and `max_node_operator_commission_share_council_adder` (inclusive)
+7. The `node_operator_commission_share_council_adder` setting SHALL be controllable by the security council without requiring a delay
+8. The security council SHOULD increment `node_operator_commission_share_council_adder` by 0.5% if the deposit pool is over half-full for the majority of a 2-week period with a constant `node_operator_commission_share + node_operator_commission_share_council_adder`
+9. `voter_share` SHALL NOT be controllable by pDAO vote and SHALL be initialized to 5%
+10. The initial pDAO settings SHALL be:
+    1. `node_operator_commission_share`: 3.5%
+    2. `node_operator_commission_share_council_adder`: 0%
+    3. `surplus_share`: 5.5%
+    4. `max_node_operator_commission_share_council_adder`: 1%
```
orangesamus, line 51: What was the rationale for lowering to only 1% of wiggle room? (Previously it was 6 seals with 0.5% each = 3% of wiggle room)?

Generally I think the less wiggle room given, the less necessary the whole addional complexity is... We are giving essentially 2 free increases at 2 weeks each vs 2 votes at ~4 weeks each. So we save 1 month of time? (Comparison is weird since I doubt we'd change back to back like that, but it helps illlustrate how it doesn't seem necessary).

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
-    3. The selected mechanism MAY be implemented in Saturn 1
-       1. If not, the revenue SHALL be held in reserve until the mechanism is implemented. In such a case, once the mechanism is implemented, the revenue SHALL be distributed in the same amount of time it took to build up, or faster.
-    4. The selected mechanism MUST be implemented in Saturn 2
-    5. When implemented, `surplus_share` SHOULD be renamed to something more specific
+6. The `node_operator_commission_share_council_adder` setting SHALL only be set to values within 0% and `max_node_operator_commission_share_council_adder` (inclusive)
+7. The `node_operator_commission_share_council_adder` setting SHALL be controllable by the security council without requiring a delay
+8. The security council SHOULD increment `node_operator_commission_share_council_adder` by 0.5% if the deposit pool is over half-full for the majority of a 2-week period with a constant `node_operator_commission_share + node_operator_commission_share_council_adder`
+9. `voter_share` SHALL NOT be controllable by pDAO vote and SHALL be initialized to 5%
```
orangesamus, line 46: Why add this comment and remove `voter_share` from "initial settings SHALL be"...

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
-    3. The selected mechanism MAY be implemented in Saturn 1
-       1. If not, the revenue SHALL be held in reserve until the mechanism is implemented. In such a case, once the mechanism is implemented, the revenue SHALL be distributed in the same amount of time it took to build up, or faster.
-    4. The selected mechanism MUST be implemented in Saturn 2
-    5. When implemented, `surplus_share` SHOULD be renamed to something more specific
+6. The `node_operator_commission_share_council_adder` setting SHALL only be set to values within 0% and `max_node_operator_commission_share_council_adder` (inclusive)
+7. The `node_operator_commission_share_council_adder` setting SHALL be controllable by the security council without requiring a delay
+8. The security council SHOULD increment `node_operator_commission_share_council_adder` by 0.5% if the deposit pool is over half-full for the majority of a 2-week period with a constant `node_operator_commission_share + node_operator_commission_share_council_adder`
```
orangesamus, line 45: Generally if we have the additionally complexity of council controlled settings, I think we might as well allow full freedom between the range (take advantage of letting it also be used to lower the NO_share if the NO queue is consistently too long). But in either case I think we either use "MAY"/move to optional heuristic section for flexibility. Or if we want it to work strictly in a similar way as the seals did, we should update this to "SHALL". Leaving it here as "SHOULD" could be confusing for the security council to decide if a specific situation is worth doing something different than "SHOULD"

RPIPs/RPIP-46.md
orangesamus, line 43: Previously there was a provision to also automatically lower "surplus_share" to account for the increased "no_share"... Did we want to add this? (Otherwise any changes to `node_operator_commission_share_council_adder` will increase the total rETH fee and pDAO vote would have to lower a different share to ensure we stay at the same rETH fee)

RPIPs/RPIP-47.md
```diff
@@ -33,8 +33,8 @@ This RPIP is part of a set of proposals motivated by a desire to rework Rocket P
   - Setting the previous `megapool delegate`'s expiration block to occur `delegate_upgrade_buffer` after the upgrade
 - A node operator SHALL be able to upgrade their active `megapool delegate` to the latest `megapool delegate` at any time (including after the active `megapool delegate`'s expiration block)
 - After the active `megapool delegate`'s expiration block:
-  - The active `megapool delegate` SHALL NOT be usable by the node operator
-  - The node operator SHALL NOT be able to claim any rewards
+  - Functions in the active megapool delegate SHALL revert when not called by a protocol contract
+  - No capital SHALL be distributed to the withdrawal address(es)
```
orangesamus, line 37:
```suggestion
  - Node operator capital SHALL be distributed to the withdrawal address(es)
```

orangesamus: I assumed by "No" you meant "NO" short for Node operator... Generally I wonder if I should sweep through all the RPIPs and change any "NO" to say node operator...

orangesamus: Also should this be made a permissionless function (in which case maybe that should go first, and then "all other functions in the active megapool delegate SHALL revert..."? Or best to just require protocol contract to distribute rewards?

RPIPs/RPIP-59.md
```diff
@@ -58,22 +56,25 @@ ETH from the deposit pool SHALL be matched with validator deposits from queues a
 - `deposit` SHALL assign deposits as described below 
 
 #### Assigning ETH from the Deposit Pool
-- As ETH enters the deposit pool, it SHALL be assigned to valdiators from the queue by sending 32 ETH to the associated megapool contract
+- ETH from the deposit pool SHALL be assigned to validators to validator at the front of the queue by sending 32 ETH to the associated megapool contract
```
orangesamus, line 59:
```suggestion
- ETH from the deposit pool SHALL be assigned to the validator at the front of the queue by sending 32 ETH to the associated megapool contract
```

RPIPs/RPIP-59.md
```diff
@@ -58,22 +56,25 @@ ETH from the deposit pool SHALL be matched with validator deposits from queues a
...
+  - rETH mints SHALL assign `ETH_deposit // 32` validators
```
orangesamus, line 60: Didn't recognize "//", looked it up and seems like it's a python thing (floor division)? Seems like solidity by default is floor division for "/"? Not sure if it's sufficient to just do "/" or if any other explicit description is needed

RPIPs/RPIP-46.md
orangesamus: Or maybe delineate “SHALL” for increment by 0.5%, but “SHOULD” for 2 week deposit pool thing

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
-   1. `node_operator_commission_share`: each NO receives this percentage of the commission from the borrowed ETH on validators they run. Unlike the remainder of the shares, this is _not_ a protocol revenue (ie, it is not socialized).
+### UARS
+1. There SHALL be the following defined shares with settings: `node_operator_commission_share`, `node_operator_commission_share_council_adder`, `voter_share`, `surplus_share`
+   1. `node_operator_commission_share + node_operator_commission_share_council_adder`: each NO receives this percentage of the commission from the borrowed ETH on validators they run. Unlike the remainder of the shares, this is _not_ a protocol revenue (ie, it is not socialized).
    2. `voter_share`: each NO receives a share of revenue based on the vote-eligible RPL staked to their megapool. The overall voter share of revenue is based on the setting, and each NO receives a proportion of that based on `vote_eligible_RPL_in_their_megapool/total_vote_eligible_RPL_in_megapools`.
    3. `surplus_share`: this share of revenue is used to distribute revenue beyond that used for protocol operation (such as the shares above)
```
knoshua, line 35: We need to also include the "adder" for surplus share. I think we also want guardrails on the "adder" parameter that make sure that it stays <= `surplus_share`
```suggestion
   3. `surplus_share - node_operator_commission_share_council_adder` - : this share of revenue is used to distribute revenue beyond that used for protocol operation (such as the shares above)
```

RPIPs/RPIP-46.md
knoshua, line 43: If no direct guardrail for adders based on `surplus_share`, could also do that on `max_node_operator_commission_share_council_adder` instead

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
+### Inflation
+1. Inflation settings SHALL be modified to retain inflation to the DAOs and eliminate inflation to NOs
+   1. `rpl.inflation.interval.rate` SHALL be set to `1000040763630249500` (1.5% per year)
+   2. Node Operators (`rocketClaimNode`) allocation SHALL be set to 0%
+   3. pDAO (`rocketClaimDAO`) allocation SHALL be set to 95%
+   4. oDAO (`rocketClaimTrustedNode`) allocation SHALL be set to 5%
+
+### Surplus revenue share vote
+Prior to the release of Saturn 1, a ranked-choice vote MUST be held to select a mechanism for surplus revenue share
+1. The choices MAY include [RPIP-45: RPL Burn](RPIP-45.md), [RPIP-50: RPL LP](RPIP-50.md), using higher `voter_share`, and/or additional options
```
knoshua, line 63: "and/or additional options" seems unnecessary given that this is started with "MAY include".

RPIPs/RPIP-46.md
knoshua: see Rationale section, line 119

RPIPs/RPIP-47.md
```diff
@@ -33,8 +33,8 @@ This RPIP is part of a set of proposals motivated by a desire to rework Rocket P
   - Setting the previous `megapool delegate`'s expiration block to occur `delegate_upgrade_buffer` after the upgrade
 - A node operator SHALL be able to upgrade their active `megapool delegate` to the latest `megapool delegate` at any time (including after the active `megapool delegate`'s expiration block)
 - After the active `megapool delegate`'s expiration block:
-  - The active `megapool delegate` SHALL NOT be usable by the node operator
-  - The node operator SHALL NOT be able to claim any rewards
+  - Functions in the active megapool delegate SHALL revert when not called by a protocol contract
```
knoshua, line 36: I don't really understand what this change is about, seems to be quite different.

RPIPs/RPIP-59.md
knoshua: I think usage of "/" in a written spec like here is likely to be understood as regular division and not floor division. We can probably even find examples of that usage of "/" in existing RPIPs. If "//" isn't understood commonly, we should say "floor division" explicitely.

knoshua (changes requested):

RPIPs/RPIP-42.md
```diff
@@ -49,10 +49,10 @@ Array indexing in this section is zero-based.
   - If `i > base_bond_array.length`: the Node Operator share before `debt` is `reduced_bond`.
   - If `i ≤ base_bond_array.length` and `i > 1`: the Node Operator share before `debt` is the amount of ETH that would bring the user's total bond down to `base_bond_array[i-2]`.
   - If `i==1`: the Node Operator share before `debt` is the amount of ETH that would bring the user's total bond down to 0 ETH.
-- Bulk validator creation/removal functions SHALL behave the same as multiple individual transactions.
+- Bulk validator creation/removal functions MAY be provided. If they are, they SHALL behave the same as multiple individual transactions.
 - If an NO has more total bonded ETH in their megapool than would be necessary based on the current settings (eg, `reduced_bond` is reduced) and they have no `debt`, it SHALL be possible to reduce their bonded ETH and receive ETH `credit` for it
 - `credit` MUST be usable to create validators in a megapool
-- `credit` MUST be usable to mint rETH to the NO's primary withdrawal address 
+- `credit` MAY be usable to mint rETH to the NO's primary withdrawal address
```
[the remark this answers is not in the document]
LongForWisdom: They aren't permanently stuck, are they? No more than someone can be permanently stuck in the current single-queue structure?

RPIPs/RPIP-43.md
```diff
@@ -171,7 +169,18 @@ Rules specifying the movement of staked RPL are as follows:
   150% of the value of total bonded ETH (from both megapools and legacy
   minipools)
   - Note that Unstaking RPL SHALL NOT be counted
-  - If this RPIP is ratified, a note SHALL be added to the "Eligibility" section of [RPIP-4](./RPIP-4.md) saying that it has been superseded by this section
+- RPIP editors SHALL modify the eligibility section of [RPIP-4](RPIP-4.md) to match the logic described in this section, and include links to both this RPIP, and the legacy version of RPIP-4.
```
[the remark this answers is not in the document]
LongForWisdom: Because RPIP-4 is used as reference, and imo shouldn't be updated to mismatch reality.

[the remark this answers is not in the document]
LongForWisdom: I would like to see RPIP-4 as Living Informational, which I think better reflects the fact its used as reference, though.

RPIPs/RPIP-47.md
LongForWisdom: It was a clarification change. It wasn't clear what the extent of 'not usable' was. So it was made more explicit: 'NO can't do anything, everything reverts unless called through another protocol contract.'

RPIPs/RPIP-59.md
LongForWisdom: Is it being used as an escape character?

orangesamus: I think we should just say "floor division" explicitly then

RPIPs/RPIP-46.md
orangesamus: We also need to consider handling this if "surplus revenue" to voter share wins...
Any changes to the adder by default affect rETH total fee, and would need pDAO vote to readjust voter share to keep rETH total fee the same (and voter_share may require 75% approval?)

RPIPs/RPIP-59.md
knoshua: > Is it being used as an escape character?

No, samsus is correct, it means floor division here.

RPIPs/RPIP-42.md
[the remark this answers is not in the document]
knoshua: I'd say more than now, since express queue increasing adds another level, but yes it is not as bad as I thought earlier.

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
+2. If higher `voter_share` is selected:
+   1. The entirety of the ["Specification taking effect with Saturn 2"](#specification-taking-effect-with-saturn-2) section below SHALL be deleted
+   2. `voter_share` SHALL become controllable by pDAO vote; however, it SHALL require a supermajority vote with at least 75% of the vote in support of any increase 
+   3. `voter_share` MAY be updated by an address in the `allowlisted_controllers` array
+3. The selected mechanism MAY be implemented in Saturn 1
+   1. If not, the revenue SHALL be held in reserve until the mechanism is implemented. In such a case, once the mechanism is implemented, the revenue SHALL be distributed in the same amount of time it took to build up, or faster.
```
knoshua, line 69:
```suggestion
   1. If not, the revenue SHALL be held in reserve until the mechanism is implemented. In such a case, once the mechanism is implemented, the revenue SHALL be distributed in the same amount of time it took to build up, or faster.
   2. If 'voter_share' is chosen, `node_operator_commission_share_council_adder` shall be subtracted from `voter_share` instead of `surplus_share`.
```

RPIPs/RPIP-46.md
```diff
@@ -28,42 +28,47 @@ A fixed percentage of RPL inflation is currently being used to fund ongoing main
...
+   2. Node Operators (`rocketClaimNode`) allocation SHALL be set to 0%
+   3. pDAO (`rocketClaimDAO`) allocation SHALL be set to 95%
+   4. oDAO (`rocketClaimTrustedNode`) allocation SHALL be set to 5%
+
+### Surplus revenue share vote
+Prior to the release of Saturn 1, a ranked-choice vote MUST be held to select a mechanism for surplus revenue share
+1. The choices MAY include [RPIP-45: RPL Burn](RPIP-45.md), [RPIP-50: RPL LP](RPIP-50.md), using higher `voter_share`, and/or additional options
+2. If higher `voter_share` is selected:
+   1. The entirety of the ["Specification taking effect with Saturn 2"](#specification-taking-effect-with-saturn-2) section below SHALL be deleted
+   2. `voter_share` SHALL become controllable by pDAO vote; however, it SHALL require a supermajority vote with at least 75% of the vote in support of any increase
```
knoshua, line 66: what is the point of having 75% when `voter_share` won? The conflict is between `surplus_share` and `voter_share`, not `no_share`. 
If we need 75% rule here, why don't we have it for changing `no_share` vs `surplus_share` from the start or for other selections of `surplus_share` ?

RPIPs/RPIP-46.md
orangesamus, line 46:
```suggestion
```

orangesamus: Not necessary if "surplus revenue to voter_share" wins, and if it doesn't then the "supermajority clause" is already spec'd elsewhere:
https://github.com/rocket-pool/RPIPs/pull/240/files/73f9eae9a3be121d7eaa569b3e1377b2d97e4faf#diff-b9a67dc438d7daecd19987642b8a1032d4d73e280ea9f4484e1856852baaf1a4R90

RPIPs/RPIP-42.md
Valdorff: No, they aren't forced to change anything unless they add more validators. They can maintain as is. Maybe:
> Since the bond curve is based on total bonded ETH when depositing, NOs adding new validators would need to add enough ETH bond to match the "enlarged bond" curve.

RPIPs/RPIP-46.md
Valdorff: Because it's not a pDAO setting -- we would not be able to vote in a new value

RPIPs/RPIP-46.md
Valdorff: Going with:
```
8. The security council SHOULD increment `node_operator_commission_share_council_adder` by 0.5% if the deposit pool is over half-full for the majority of a 2-week period with a constant `node_operator_commission_share + node_operator_commission_share_council_adder`
    1. The security council SHALL NOT otherwise change `node_operator_commission_share_council_adder`
```

RPIPs/RPIP-46.md
Valdorff: Yes, that's my b.

Valdorff: Changing it above:

> 3. `surplus_share -  node_operator_commission_share_council_adder`: this share of revenue is used to distribute revenue beyond that used for protocol operation (such as the shares above)

RPIPs/RPIP-47.md
Valdorff: Quite the opposite -- I meant that no capital (aka zero capital) SHALL be distributed

RPIPs/RPIP-59.md
Valdorff: Going with 
> - rETH mints SHALL assign `floor(ETH_deposit / 32)` validators

RPIPs/RPIP-43.md
```diff
@@ -99,18 +100,15 @@ into shares is defined in [RPIP-46](RPIP-46.md).
...
   - When called, capital borrowed from the protocol that has been released from
     exited validators SHALL be sent to the rETH contract.
     - If the capital is insufficient to repay the protocol, the shortfall SHALL be added to `debt`
   - When called while the megapool has `debt`, the remaining capital from exited validators SHALL first be used to pay off `debt`
   - When called, the remaining capital SHALL then be held in the megapool as unclaimed node operator funds
-  - This function SHALL allow any user to call it following a mandatory time
-    delay configurable by the pDAO. The delay SHALL be initialized by a
-    `startUserDistribute` function. After the delay is complete, any user may
-    then call the capital distribution fuction.
+  - This function SHALL be permissionless following a mandatory time delay configurable by the pDAO. The delay SHALL be initialized by a `startUserDistribute` function. After the delay is complete, any user may then call the capital distribution function.
```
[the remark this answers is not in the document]
Valdorff: You're right - should be windowed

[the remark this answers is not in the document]
Valdorff: >   - This function SHALL support permissionless use, with additional restrictions:
    - There SHALL be pDAO settings called `userDistributeDelay` (initially set to 30 days) and `userDistributeWindow` (initially set to 60 days)
    - A permissionless `startUserDistribute` function SHALL record the time when called, as long as there is no previously recorded time _or_ the previously recorded time was longer than `userDistributeWindow` ago
    - The capital distribution function may be called permissionlessly if `userDistributeDelay` has passed since the `startUserDistribute` call, but `userDistributeWindow` has not yet passed since the `startUserDistribute` call

How's this?

RPIPs/RPIP-43.md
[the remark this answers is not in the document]
Valdorff: ? RPIP-4 Isn't just informational imo. It describes our active system of snapshot governance, which isn't going away.

RPIPs/RPIP-46.md
Valdorff: Added the subtraction to surplus share above. Then I added the guardrail as:

> 6. The `node_operator_commission_share_council_adder` setting SHALL only allow values where:
   1. 0% ≤ `node_operator_commission_share_council_adder` ≤ `max_node_operator_commission_share_council_adder` 
   2. `node_operator_commission_share_council_adder` ≤ `surplus_share`

RPIPs/RPIP-46.md
Valdorff: went another route above

RPIPs/RPIP-46.md
Valdorff: Yep - deleting that line

RPIPs/RPIP-46.md
Valdorff: Used ~your text, but within the "If higher `voter_share` is selected:" bullet instead

RPIPs/RPIP-46.md
Valdorff: Added within "If higher `voter_share` is selected:"
> 4. `voter_share` SHALL be a pDAO setting (and thus the specification item about it not being controllable by pDAO vote MUST also be removed)

RPIPs/RPIP-46.md
```diff
@@ -32,17 +32,20 @@ This RPIP is part of a set of proposals motivated by a desire to rework Rocket P
 1. There SHALL be the following defined shares with settings: `node_operator_commission_share`, `node_operator_commission_share_council_adder`, `voter_share`, `surplus_share`
    1. `node_operator_commission_share + node_operator_commission_share_council_adder`: each NO receives this percentage of the commission from the borrowed ETH on validators they run. Unlike the remainder of the shares, this is _not_ a protocol revenue (ie, it is not socialized).
    2. `voter_share`: each NO receives a share of revenue based on the vote-eligible RPL staked to their megapool. The overall voter share of revenue is based on the setting, and each NO receives a proportion of that based on `vote_eligible_RPL_in_their_megapool/total_vote_eligible_RPL_in_megapools`.
-   3. `surplus_share`: this share of revenue is used to distribute revenue beyond that used for protocol operation (such as the shares above)
+   3. `surplus_share -  node_operator_commission_share_council_adder`: this share of revenue is used to distribute revenue beyond that used for protocol operation (such as the shares above)
 2. `reth_commission` SHALL be defined as the sum of all defined shares that have settings
```
orangesamus, line 36: > 1. There SHALL be the following defined shares with settings: `node_operator_commission_share`, `node_operator_commission_share_council_adder`, `voter_share`, `surplus_share`
...
> `reth_commission` SHALL be defined as the sum of all defined shares that have settings

Hmm does `node_operator_commission_share_council_adder` make this not true anymore? Maybe it should say:
> `reth_commission` SHALL be defined as the sum of `node_operator_commission_share`, `voter_share`, and `surplus_share`

or something else?

RPIPs/RPIP-46.md
orangesamus: Makes sense...

PR #241, opened 2024-07-10
[TOKENOMICS, FOR REVIEW] - RPIP-47: Forced Delegate Upgrades

LongForWisdom:
Rewrote RPIP-47 to:
* Remove recovery delegate and switch to direct-upgrade-to-latest for scope reasons.
* Actually specify for the enforcement of the upgrade grace period.


RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
-  - Final distribution of rewards and capital 
-- The `megapool recovery delegate` SHOULD be updated only rarely, and only to support changes to the Ethereum execution layer exit process, or major changes in the Rocket Pool protocol. Simultaneous upgrades to both the `megapool recovery delegate` and the latest `megapool delegate` SHOULD be avoided if possible.
+
+### Upgrade Buffer Enforcement
+- Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
+- Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value.
+  - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days.
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be:
+  - For the pDAO, oDAO, or Security Council to set its value to now + its `upgrade_buffer`
+  - For the pDAO, oDAO, or Security Council to set its value to "no expiration"
```
LongForWisdom, line 36: Taking suggestions for who should be able to do these. Currently all just because it was simple.

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
-- The `megapool recovery delegate` SHALL have the following capabilities:
-  - Force exiting ALL validators under the megapool
-    - In the Saturn 1 release, this MAY be a non-functional stub
-  - Removal of any pending validators from the Rocket Pool queue
-  - Processing of voluntary exit proofs
-  - Collection of any penalties or deficits
-  - Final distribution of rewards and capital 
-- The `megapool recovery delegate` SHOULD be updated only rarely, and only to support changes to the Ethereum execution layer exit process, or major changes in the Rocket Pool protocol. Simultaneous upgrades to both the `megapool recovery delegate` and the latest `megapool delegate` SHOULD be avoided if possible.
+
+### Upgrade Buffer Enforcement
```
LongForWisdom, line 30: There's an argument that it makes sense to let the Security Council force upgrade ignoring expiry to handle bugs. In that case it might still be worth having this on-chain.

If people are uncomfortable with not having an out in case of bugs / exploits and want the pDAO + oDAO to be able to upgrade whenever, then all this should just move off-chain.

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
+- Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
+- Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value.
+  - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days.
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be:
+  - For the pDAO, oDAO, or Security Council to set its value to now + its `upgrade_buffer`
```
LongForWisdom, line 35: This could be used to preempt upgrades to functionally skip the buffer. Possible fix to this is to require a target address to upgrade to alongside the buffer start. This would mean the contract would have to exist prior to a reset.

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
+  - For the pDAO, oDAO, or Security Council to set its value to "no expiration"
+
+### Upgrade
+- When a new `megapool delegate` is released, a contract upgrade SHOULD update all previous `megapool delegate`s `expiration_block`'s to now + their `upgrade_buffer`s
```
knoshua, line 39: a contract upgrade updating the `expiration_block` conflicts with who gets to change it as specced above.
I think allowing network contracts to do modify it makes sense. Adding of network contracts is gated by oDAO.

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
+- A node operator SHALL be able to upgrade their active `megapool delegate` to the latest `megapool delegate` at any time
+- After an active `megapool delegate`'s expiration block has passed:
+  - The expired `megapool delegate` SHALL NOT be usable by the node operator
+  - The node operator SHALL NOT be able to claim any rewards using the expired `megapool delegate`
```
knoshua, line 43: Do we want to match the updated wording from Val's recent PR here?

```suggestion
  - Functions in the expired megapool delegate SHALL revert when not called by a protocol contract
  - No capital SHALL be distributed to the withdrawal address(es)```

File: RPIPs/RPIP-47.md [hunk 6]
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
+  - The expired `megapool delegate` SHALL be permissionlessly upgradeable to the latest `megapool delegate`
```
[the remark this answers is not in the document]
knoshua: Definitely should include the node operator. Not sure if more than that is necessary or desirable.

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,34 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
+- Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
+- Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value.
+  - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days.
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be:
+  - For the pDAO, oDAO, or Security Council to set its value to now + its `upgrade_buffer`
+  - For the pDAO, oDAO, or Security Council to set its value to "no expiration"
+
+### Upgrade
+- When a new `megapool delegate` is released, a contract upgrade SHOULD update all previous `megapool delegate`s `expiration_block`'s to now + their `upgrade_buffer`s
+- A node operator SHALL be able to upgrade their active `megapool delegate` to the latest `megapool delegate` at any time
```
knoshua, line 40: Currently, with minipool delegates, node operators get to set a flag to true/false to always use latest. Not sure if there's been discussing about that kind of functionality for megapools.

```suggestion
- A node operator SHALL be able to upgrade their active `megapool delegate` to the latest `megapool delegate` at any time
- A node operator SHALL be able to choose to always and automatically use the latest `megapool delegate`
```

RPIPs/RPIP-47.md
knoshua: I think simply setting `expiration_block` without having a new delegate would raise all kinds of alarm bells. I'm not sure that requiring a target address (which might be bogus, upgradeable or just not yet deployed) adds much.

RPIPs/RPIP-47.md
knoshua: What's the scenario where we'd want to set it back to "no expiration"? Can we just make this "network contract" (see also below)? oDAO would have to vote on it, Security Council could veto, pDAO has authority over upgrades.

RPIPs/RPIP-47.md
LongForWisdom: Yes, I'll merge that in, didn't want to double up on it before it was confirmed.

RPIPs/RPIP-47.md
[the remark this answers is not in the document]
LongForWisdom: The NO is always able to upgrade, see ln40.

RPIPs/RPIP-47.md
LongForWisdom: Back to no expiration would be if the new delegate was discovered to be bugged, etc. In a way where it wasn't easy to fix before the previous one expired. Expirations may not always be 120 day in the future.

LongForWisdom: I'll update both to network contract.

RPIPs/RPIP-47.md
LongForWisdom: I wasn't aware, I can add the line, seems sensible to maintain functional parity.

LongForWisdom: On balance, feels like it might be scope creep, so I'm going to make it a MAY

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,35 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
-- After the active `megapool delegate`'s expiration block:
-  - Functions in the active megapool delegate SHALL revert when not called by a protocol contract
+
+### Upgrade Buffer Enforcement
+- Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
+- Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value
+  - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be:
+  - Any protocol contract may set its value to now + its `upgrade_buffer`
+  - Any protocol contract may set its value to "no expiration"
```
Valdorff, line 36: what's the use case here?

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,35 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
-- The `megapool recovery delegate` for a megapool SHALL be usable by any address if the current block is greater than the active `megapool delegate`'s expiration block
-- The `megapool recovery delegate` SHALL have the following capabilities:
-  - Force exiting ALL validators under the megapool
-    - In the Saturn 1 release, this MAY be a non-functional stub
-  - Removal of any pending validators from the Rocket Pool queue
-  - Processing of voluntary exit proofs
-  - Collection of any penalties or deficits
-  - Final distribution of rewards and capital 
-- The `megapool recovery delegate` SHOULD be updated only rarely, and only to support changes to the Ethereum execution layer exit process, or major changes in the Rocket Pool protocol. Simultaneous upgrades to both the `megapool recovery delegate` and the latest `megapool delegate` SHOULD be avoided if possible.
+  - The expired `megapool delegate` SHALL be permissionlessly upgradeable to the latest `megapool delegate`
```
Valdorff, line 45: I think we can have this simply be:

```
- After an active `megapool delegate`'s expiration block has passed, the expired `megapool delegate` SHALL be permissionlessly upgradeable to the latest `megapool delegate`
```

There's no need to break the existing delegate (and thus make the logic and spend the gas to check for expiry) if we can simply upgrade folks. We can have a line in security considerations around incentivizing users to permissionlessly upgrade, eg with GMC funds - no need to define that further right now.

RPIPs/RPIP-47.md
Valdorff: I think I'm comfortable as written

RPIPs/RPIP-47.md
LongForWisdom: Lol Knoshua asked the same thing and I resolved it. It's mainly intended as a fallback. If we discover something is wrong that can't be quickly fixed, it might make more sense to set it back to no-expire rather than pushing it 120 days.

RPIPs/RPIP-47.md
LongForWisdom: You're right. I will update on Monday.

RPIPs/RPIP-47.md
Valdorff: I don't think I like that at all. We could just redeploy a copy of the old thing as latest. Then it would be "no expiration" and also properly work for "update to latest". I recommend removing this.

LongForWisdom: I don't think I agree. Redeploying a copy is more work and forces someone to pay gas to move people from one identical contract to another. The contracts need to support no-expiration status anyway as the initial value. Can you elaborate on why you don't want this?

Valdorff: The example you gave is:

- GoodDelegateA has been deployed a while
- BadDelegateB is deployed; GoodDelegateA gets an expiration block
- We realize and say "oops". We set GoodDelegateA's expiration block to nothing

In this new state we have:
- Two non-expiring delegates that people might be on
- No way for people on BadDelegateB to get to a GoodDelegate
- A trap -- anyone that upgrades ends up on a BadDelegate
- Another trap -- anyone that makes a new node will use the latest... which is BadDelegateB

-----

My proposal would instead be to deploy GoodDelegateC (which may be a copy of A or not; but let's say it is for now).

In that state we have:
- a single non-expiring delegate (C)
- a way for people on BadDelegate to get to a GoodDelegate
- no traps

Yes, people may  need to waste gas going from A to C, which is the same. They only need to do this if (A) A is expiring, (B) the latest delegate isn't C (cuz being force upgraded to a copy isn't at all damaging), (C) they're not wanting to upgrade to the latest delegate (ditto)

LongForWisdom: Discussed some with Val on discord. TL;DR is that if we were to keep it, it would be to handle unforeseen issues that we can't think of. Any feasible issue should be resolvable by just redeploying a known-good delegate as the latest, which is slightly more painful. In most cases we can think of, you'd need to replace the delegate anyway to prevent voluntary upgrades to a bad contract.

LongForWisdom: So line removed and comment resolved.

RPIPs/RPIP-47.md
darrenlangley: From my perspective, it is fine because, in the case of a bug, we can always deploy another delegate and people can roll forward. We couldn't force people over to the new delegate within the grace period but that is part of the social contract and so it is the trade off that we are making.

RPIPs/RPIP-47.md
```diff
@@ -26,34 +26,30 @@ Additionally, the pDAO must have the ability to effect change in ways that benef
...
-- The `megapool recovery delegate` SHOULD be updated only rarely, and only to support changes to the Ethereum execution layer exit process, or major changes in the Rocket Pool protocol. Simultaneous upgrades to both the `megapool recovery delegate` and the latest `megapool delegate` SHOULD be avoided if possible.
+
+### Upgrade Buffer Enforcement
+- Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
+- Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value
+  - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to now + its `upgrade_buffer`
+
+### Upgrade
+- When a new `megapool delegate` is released, a protocol contract SHOULD update all previous `megapool delegate`s `expiration_block`'s to now + their `upgrade_buffer`s
```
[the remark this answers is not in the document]
LongForWisdom: Like, the whole point of enforcing it here is that its enforcing it _against_ changes in the RP governance process. 

In usual practice, I can't see that this is even going to have much of an impact. I strongly doubt we're ever going to be releasing new and different delegates within the 120 day limit. 

So in usual circumstances this has zero effect, as there is only ever likely to be a single active delegate version. 
However, it also gives NO's some protection against unusual circumstances.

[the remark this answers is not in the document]
Valdorff: I think you're seeing this differently than I. It's not a grace period at all to me. It's an observation period to see the delegate operating in the real world and getting lindy. 

So the situation is:
- I'm on delegate X on day 0
- Delegate Y is released
- I begin observing, with up to 120 days available
- After 60 days, delegate Z is proposed

My options are now:
- Decide I'm comfortable enough to go with delegate Y before Z is live
  - This would give me until day 180 to make a decision on Z
- Let Z go live; before day 120:
  - Decide I'm comfortable with Z
  - Decide I'm not comfortable and exit

This all seems ok to me. In edge cases a Node Operator may feel the need to exit, but they aren't forced into anything ever.

----
Your variant is reasonably similar:
- I'm on delegate X on day 0
- Delegate Y is released
- I begin observing, with up to 120 days available
- After 60 days, delegate Z is proposed

My options are now:
- Decide I'm comfortable enough to go with delegate Y before Z is live
  - This would give me until 180 to make a decision on Z
- Let Z go live; before day **180** <--- this is the only change:
  - Decide I'm comfortable with Z
  - Decide I'm not comfortable and exit
----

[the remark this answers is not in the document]
LongForWisdom: Hmm. I agree there is not much difference in the usual case. But I do feel the 'all active' version is preferable for the unusual case with little downside.

Consider the following, as a small alteration to your first sccenario:

I'm on delegate X on day 0
* Delegate Y is released
* I begin observing, with up to 120 days available
* After 119 days, protocol is upgraded to delegate Z (I know this is unlikely in practice)

My options are:
* Decide I don't like Y, and bail with plenty of time (all good here)
* Decide I'm comfortable enough to upgrade to delegate Y before Z is live, (all good here too).
* Decide I'm comfortable with delegate Y, but I'm lazy, so I don't manually upgrade. When Z comes in on day 119, I have effectively no time to evaluate, and I'm moved to Z, possibly before I even notice Z has been proposed. **This is very bad.**

---

It just seems very low cost to me to prevent the third scenario by resetting the grace/observation period. Yes it's unlikely to occur, but resetting active delegate versions (when it's also very unlikely they'll be more than one at a time) really seems like an easy way to just let NO's never need to worry about this.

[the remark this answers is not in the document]
LongForWisdom: I'll update the wording to some variation of 'delegate versions currently in use', tomorrow, but I do feel that's a better version than just updating the last delegate.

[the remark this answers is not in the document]
darrenlangley: Hey, yes we agree with you @LongForWisdom. Some variation of 'delegate versions currently in use' or 'active' would be fine. We were a little worried about implementation but have review and are confident it can be achieved.

LongForWisdom:
Okay, changes put in to tighten up language around all previous delegate as discussed. Should be good to merge?

PR #243, opened 2024-07-12
[TOKENOMICS, FOR REVIEW] RPIP-60: Protocol Upgrade Guardrails

LongForWisdom:
Update for RPIP-60. Main changes:
* Remove fast-upgrades as a scope reduction.
* Add security considerations and rationale sections.
* Slightly relax requirements around the veto explanation document. 
* Rename 'default_upgrade_delay' -> 'upgrade_delay', and clarify it's a pDAO setting.


Valdorff:
Looks good to me

Valdorff:
I can't put this in as a suggestion in git cuz it bridges a deleted line, but here's some expanded Security and Rational sections. Use in toto or in part as you wish.


```
## Security Considerations

- The security council veto means it would require both a rogue oDAO and a rogue security council to pass a corrupt proposal.
- The security council veto means that a rogue security council could block a valid proposal.
  - This is mitigated by the pDAO's ability to replace the security council. While the valid proposal could be blocked once, it could not be blocked indefinitely.
- The veto process can itself be modified by contract upgrades. However, if a corrupt proposal tries to change the veto process, that can be defended against by using the veto process.
- Emergency response settings changes should be taken directly by the security council (see [RPIP-33](RPIP-33.md)) rather than use the regular governance process.
- The fastest possible upgrade speed is significantly slowed from the previous state. It used to take `proposal.vote.delay.time` (currently 1 week) and will now take an additional `upgrade_delay` (initially set to 3 weeks). This means that important hotfixes could take a lot longer. It may also mean that emergency pauses (eg, disabling rETH deposits) could last longer until a hotfix takes effect.
- The fastest possible upgrade speed exactly matches the RPL `unstaking_period` defined in [RPIP-30](RPIP-30.md). 

## Rationale

**Veto Explanation Document**  
The explanation document is a mechanism intended to help the pDAO hold the security council to account for their decision-making around vetos. It gives more information as to the reasoning behind a veto decision, and should help to inform pDAO decision-making post veto. In an ideal world the document accompanies the veto, the 24 hour grace period is intended to cover situations in which the security council must act within a limited time window.

**Upgrade Delay Setting**
Due to the considerations around slowing down hotfixes [described above](#security-considerations), the initial upgrade delay setting is attempting to be the smallest that allows "enough" time for participants to exit the system ahead of a proposal if they deem that appropriate.

Any upgrade will have `proposal.vote.delay.time` (currently 1 week) and `upgrade_delay` (initially set to 3 weeks). This is technically _just_ enough to match the RPL `unstaking_period` (currently 28 days). In normal realistic governance, there will be more time than that: there is a leadup period to sentiment poll (currently at least 1 week), there is a pDAO vote period (currently at least 2 weeks), there's a human action to raise the oDAO proposal, and there are many human actions to place oDAO votes after `proposal.vote.delay.time`. These delays provide some human breathing room to allow participants to exit ahead of a proposal.
```

Valdorff:
One more bullet for secuiryt considerations:
> - The security council veto means that if pDAO governance were captured, protocol upgrades could be stopped at will (by installing a compromised security council). The oDAO would be unable to pass any contract upgrades (without the approval of the compromised security council), and the only way back would be to recapture pDAO governance. While this means there is a real tradeoff, this "inability to upgrade" is seen as less problematic than the "ability to pass any upgrade" that a rogue oDAO would have without the veto.

LongForWisdom:
@Valdorff incorporated the above, more or less verbatim.

PR #244, opened 2024-07-12
3 tx deposits

knoshua:
meant for discussion


LongForWisdom:
Note, we're sticking with 2TX as part of initial batch of RPIPs, and are explicitly leaving room open to switch to 3TX later if it seems better.

PR #246, opened 2024-07-16
[TOKENOMICS, FOR REVIEW] Updates to RPIP-49 to better track the current state of the rework.

LongForWisdom:
Would appreciate comment from @Valdorff to ensure 'still-to-ratify' list meets expectations.


RPIPs/RPIP-49.md
```diff
@@ -137,10 +145,10 @@ The below is generally agreed to be the steps to be completed before we can cons
 3. **Done** - Create high-level explanations and informational material for the full proposal for consumption by the wider community.
 4. **Done** - Make a concerted effort to gather feedback via the forum from the wider community.
 5. **Active** - Update the proposal and specifications as needed taking into account wider community feedback.
-6. Run a forum temperature check vote on the full proposal (bar value capture mechanism).
-7. Run a snapshot vote on the full proposal (bar value capture mechanism).
-8. Run a snapshot vote on the value capture mechanism once outstanding blockers are cleared.
-9. Update the proposal to incorporate the value capture mechanism vote result.
+6. Run a forum temperature check vote on the rework package (bar the 'still-to-ratify' list above).
+7. Run a snapshot vote oon the rework package (bar the 'still-to-ratify' list above).
```
orangesamus, line 149:
```suggestion
7. Run a snapshot vote on the rework package (bar the 'still-to-ratify' list above).
```

RPIPs/RPIP-49.md
```diff
@@ -137,10 +145,10 @@ The below is generally agreed to be the steps to be completed before we can cons
...
+8. Run a snapshot votes on the 'still-to-ratify' list as outstanding blockers are cleared.
```
orangesamus, line 150:
```suggestion
8. Run a snapshot vote on the 'still-to-ratify' list as outstanding blockers are cleared.
```

orangesamus: I think the expectation to limit unnecessary voting will be to have a primary vote on the rework package (in the short term future... "step 7"), and then run a snapshot vote altogether at a later date after we don't expect any more changes to be needed. Maybe the "surplus share" vote should be it's own thing though since it is more of an opinionated vote with larger implications (compared to just implementation nuance changes that the team expressed desire to have flexibility to change at a later date as they actually begin development).

You could change to:
> Run snapshot votes on the 'still-to-ratify' list as outstanding blockers are cleared.

or
> Run snapshot vote(s) on the 'still-to-ratify' list as outstanding blockers are cleared.

Any of those are fine I think

Valdorff: I like the latter samus wording -- gives flexibility to be one or multiple

Valdorff, line 150:
```
7. Run a snapshot vote on the rework package as Living RPIPs, acknowledging the existence of the [Still To Ratfiy Prior to Saturn 1](#still-to-ratify-prior-to-saturn-1) list above.
8. Run one or more snapshot vote(s) as blockers are cleared from the 'still-to-ratify' list. The end state will include no remaining blockers and the status of the rework RPIPs set to Final.
```

RPIPs/RPIP-49.md
```diff
@@ -120,14 +121,21 @@ The tokenomics rework package will likely be split between two protocol upgrades
...
 
 Current efforts are primarily focused on:
-1. Updating and polishing the RPIP specifications in response to feedback.
-2. Attempting to reduce the scope of the required implementation work to improve time-to-market without overly impacting contents.
+1. Confirming Core Team is happy with RPIP specification contents and making minor adjustments as needed.
+2. Polishing the RPIP specifications ready for vote.
 
 A maintained list of open tasks is available via google sheet [here](https://docs.google.com/spreadsheets/d/1KDTeFnNl3XLFO37upti6NbT2p2GDYJ4GKH4aJ51gQZA/edit?pli=1#gid=725857744). 
 
+### Still To Ratfiy Prior to Saturn 1
```
Valdorff, line 132: Spelling: Ratify

RPIPs/RPIP-49.md
```diff
@@ -120,14 +121,21 @@ The tokenomics rework package will likely be split between two protocol upgrades
...
+These items are to be considered flexible until ratified explicitly, rather than ratified alongside the tokenomics rework package as a whole. 
+
+* **Saturn 2 Surplus Share Strategy** - Decide and ratify how to manage surplus share in Saturn 2 (we want to decide this before Saturn 1 launches to avoid the incumbent advantage for the voter_share option.)
+* **Deposit Strategy** - Decide whether to take a 2TX or 3TX deposit strategy for validator creation and ratify choice.
+* **Maximum Megapool ETH Penalty** - Decide and ratify a non-placeholder value for the maximum penalty, likely needs community research to find a good value.
+
```
Valdorff, line 138: - `vote_eligible_target_min` and `vote_eligible_target_max`
This shouldn't block voting on the main thing, but we need to choose em (and actually we only need to choose them if voter_share doesn't win the surplus share strategy vote)... could be a sub-bullet under that then?

RPIPs/RPIP-49.md
LongForWisdom: Sigh, should have run grammarly. Thanks Samus.

RPIPs/RPIP-49.md
LongForWisdom: Added. None of these are blocking on the main vote.

RPIPs/RPIP-49.md
LongForWisdom: Went with this, thanks both.

RPIPs/RPIP-49.md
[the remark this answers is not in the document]
LongForWisdom: This isn't a single item, so I'm not sure it belongs in this list. We can add it in somewhere if you think it's important it be in this document, but so far I'd planned to just include it in the text for both forum and snapshot votes.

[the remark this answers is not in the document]
LongForWisdom: I'll add some note at the bottom I guess.

RPIPs/RPIP-49.md
```diff
@@ -120,14 +121,21 @@ The tokenomics rework package will likely be split between two protocol upgrades
...
+2. Polishing the RPIP specifications ready for vote.
 
 A maintained list of open tasks is available via google sheet [here](https://docs.google.com/spreadsheets/d/1KDTeFnNl3XLFO37upti6NbT2p2GDYJ4GKH4aJ51gQZA/edit?pli=1#gid=725857744). 
 
+### Still To Ratfiy Prior to Saturn 1
+These items are to be considered flexible until ratified explicitly, rather than ratified alongside the tokenomics rework package as a whole. 
+
+* **Saturn 2 Surplus Share Strategy** - Decide and ratify how to manage surplus share in Saturn 2 (we want to decide this before Saturn 1 launches to avoid the incumbent advantage for the voter_share option.)
+* **Deposit Strategy** - Decide whether to take a 2TX or 3TX deposit strategy for validator creation and ratify choice.
+* **Maximum Megapool ETH Penalty** - Decide and ratify a non-placeholder value for the maximum penalty, likely needs community research to find a good value.
```
Valdorff, line 138: I think knoshua has convinced me this is too narrow for our current confidence level. Maybe more like 
> **Penalty System** - Research, draft and ratify a penalty system. Right now, there is an idea written, with a missing Maximum ETH Penalty. There has been some discussion around, eg, making this a 2 step "freeze and then determine penalty" process. Work is needed, but this doesn't change the overall proposal much at all.

LongForWisdom: Have added it in an abbreviated form, I don't think we need to go into significant detail here.

PR #249, opened 2024-07-18
Security Considerations RPIP-59

knoshua:
- added Security Considerations and moved content of Considerations section into it
- removed one unnecessary storage requirement from `deposit`
- clarified what validating in `deposit` means
- clarified assignment as part of `deposit`
- fixed some formatting


Valdorff:
Looks great :)

RPIPs/RPIP-59.md
```diff
@@ -91,11 +92,26 @@ ETH from the deposit pool SHALL be matched with validator deposits from queues a
...
+  - With Saturn 2, we update `reduced_bond` to 1.5 ETH, which further disincentivizes sock puppets
+- Setting `express_queue_rate` to 0 disables the express queue and people currently in it are stuck indefinitely. Similarly, setting `express_queue_rate` to a very high value effectively disables the `standard_queue`. However, node operators would be able to exit the queue for `credit`.
+
+### Deposit Mechanics
+- Validation of provided data in `deposit` matches the validation that the beacon chain deposit contract does and therefore ensures that the `prestake` using the data will succeed
+- Disabling social assignments increases the likelihood of ETH accumulating in the depoist pool while validators are waiting in queue. But the node operators in queue that would be assigned ETH have a direct incentive to execute the assignments. In addition, we recommend that the pDAO funds (for example through the GMC) development and running of assignment bots that assign ETH in the deposit pool to the queue at a reasonable gas price.
```
orangesamus, line 108:
```suggestion
- Disabling social assignments increases the likelihood of ETH accumulating in the deposit pool while validators are waiting in the queue. But the node operators in the queue that would be assigned ETH have a direct incentive to execute the assignments. In addition, we recommend that the pDAO funds (for example through the GMC) development and execution of assignment bots that assign ETH in the deposit pool to the queue at a reasonable gas price.
```

orangesamus: feel free to just take the "depoist" typo fix, or the rest as well

RPIPs/RPIP-59.md
```diff
@@ -68,7 +68,7 @@ ETH from the deposit pool SHALL be matched with validator deposits from queues a
 - If the beacon chain stake is invalid, the validator SHALL be scrubbed 
 - `stake` SHALL stake the remaining 31 ETH to the beacon chain to make a complete validator
 - If `stake` is not called within `time_before_dissolve` after the ETH was assigned, the validator SHALL be dissolved, returning the unstaked balance to the deposit pool
-  - If a validator is dissolved the bonded value SHALL be recoverable. This MAY require further action from the node operator. This MAY temporarily require additional ETH from the node operator
+  - If a validator is dissolved the bonded value SHALL be recoverable. This MAY require further action from the node operator. This MAY temporarily require additional ETH from the node operator.
```
orangesamus, line 71: Not sure what convention (if any) we are using for periods, noticed elsewhere some sentences do/don't have periods. I didn't make a suggestion commit adding periods everywhere else since it seems unnecessary.

knoshua: it seems like, convention seems to be no period on bullet points. But I have some bullet points with multiple sentences (like here) and then it feels weird to have a period in the middle but not at the end.

LongForWisdom: I feel like we should just change the convention to have periods on all bullets, but agree this is better than having periods in the middle and not end.

Valdorff: Most style guides accept _either_ no periods on a bulleted list, or periods on every bullet. But also - I'm hitting the merge button. The relevant info is there and not confusing.

PR #257, opened 2024-07-22
Fixed the calculation of ETH addition / refund for validator creation / exit

sckuzzle:
The calculation changes fixes bugs that could arise due to changing bond requirements after a node has already created validators.


Valdorff:
Looks good 🙏

PR #266, opened 2024-07-22
Explainers final cleanup

orangesamus:
Fixed some calculations and cleaned up to consistent formatting

Made one stylistic choice: switching proposal using current LEB8 allocation, and ETH-Only LEB8 allocation. 
This helps make the comparison to existing LEB8 tokenomics more clear.

See example final version here:
<https://orangesamus.github.io/RPIPs/tokenomics-2024>


tokenomics-explainers/003-rework-foundation.md
```diff
@@ -161,12 +161,12 @@ The liquidity pool is a weighted pool with 90% RPL weight and 10% rETH weight. T
...
+In this case, no share of ETH revenue is used to buy RPL (which means unstaked RPL does not directly benefit from the ETH revenue). Instead, the equivalent share of ETH revenue is simply used to have a larger `voter_share` which is distributed according to vote-eligible RPL staked to megapools. This means only node operators are natively able to benefit from RPL’s value capture directly. This has some positives (direct ETH received instead of indirect value captured; avoids some complexity in implementing one of the above options) and some negatives (much smaller total addressable market). Note that it may become possible to benefit without being a node operator via abstractions (eg, Nodeset’s proposed xRPL).
```
orangesamus, line 164:
```suggestion
In this case, no share of ETH revenue is used to buy RPL (which means unstaked RPL does not directly benefit from the ETH revenue). Instead, the equivalent share of ETH revenue is simply used to have a larger `voter_share` which is distributed according to vote-eligible RPL staked to megapools. This means only node operators are natively able to benefit from RPL’s value capture directly. This has some positives (direct ETH received instead of indirect value captured; avoids some complexity in implementing one of the above options) and some negatives (much smaller total addressable market). Note that it may become possible to benefit without being a node operator via abstractions (eg, NodeSet’s proposed xRPL).
```

PR #272, opened 2024-07-28
[COMMENTS ONLY]

orangesamus:
Included some comments/questions that came up for me when going through the final checklist... If any of them should be updated I can add to PR271


RPIPs/RPIP-43.md
```diff
@@ -32,7 +32,7 @@ This specification introduces a type of smart contract called a "megapool" that
 SHALL serve as the target of beacon chain withdrawal credentials for Rocket Pool
 validators deposited after this RPIP is implemented.
 
-- A node operator SHALL be able to deploy at most one megapool contract
+- A node operator SHALL be able to deploy at most one megapool contract per node
```
orangesamus, line 35: Technically we don't have visibility if the same node operator opens multiple megapool contracts on different nodes, so should this more clearly state "per node", or "A node SHALL be able to..."

Maybe that muddies the descriptions everywhere else though and this should be ignored

RPIPs/RPIP-44.md
```diff
@@ -22,7 +22,7 @@ There is at least one additional use case not addressed in this RPIP, which is t
 
 Execution Layer Triggerable Exits are motivated by the need to combat MEV theft by malicious Rocket Pool validators. Currently, the protocol has limited recourse against such offending validators and the proposed functionality helps to change this. It's important to continue to pursue such improvements to ensure rETH delivers the advertised APY and competes favorably with other liquid staking tokens. 
 
-A secondary motivation is to improve the node operator user experience, the ability to request exit of their validators via the protocol is an improvement to the current experience.
+A secondary motivation is to improve the node operator user experience, the ability to request exit of their validators via the protocol is an improvement to the current experience (e.g., in the case of abandonment a forced exit by the protocol would be preferable to continuous leakage penalties).
```
orangesamus, line 25: Is abandonment/incapacitation examples of what this sentence was trying to hint at? I wasn't sure, and maybe including further description/example would help.

RPIPs/RPIP-44.md
```diff
@@ -59,6 +59,7 @@ The exit functionality for a megapool is specified as follows:
       - It MUST NOT be possible for accounts other than the node operator to exit more validators than needed to reduce `deficit` below `exit_deficit`
 - The protocol SHALL use the `withdrawn`, `credit`, and `received` balances to pay off `debt` prior to taking action on exits
 - The protocol MAY use unstaking and staked RPL to decrease `debt` by the corresponding amount prior to taking action on exits
+- The protocol MAY use unclaimed rewards to decrease `debt` by the corresponding amount prior to taking action on exits
```
orangesamus, line 62: Should "unclaimed rewards" also be listed? Maybe needs further descriptions to if so

RPIPs/RPIP-60.md
```diff
@@ -54,7 +54,7 @@ The explanation document is a mechanism intended to help the pDAO hold the secur
...
+Any upgrade will have `proposal.vote.delay.time`<-{I'm guessing this is a setting defined somewhere else since it says "currently 1 week"... should any other context be listed for this spec?} (currently 1 week) and `upgrade_delay` (initially set to 1 week). This is likely enough time to exit validators and withdraw ETH (barring an extreme exit queue). In normal circumstances, there will be more time available to exit. There is a leadup period to the sentiment poll (currently at least 1 week), there is a pDAO vote period (currently at least 2 weeks), there's often dev time (which can be large), there's a human action required to raise the oDAO proposal, and there are many human actions to place oDAO votes after `proposal.vote.delay.time`. These delays should provide significant breathing room to allow participants to exit their ETH positions ahead of a proposal, and may even provide enough time to exit RPL positions despite the 28-day unstaking period. RPL exiting was not heavily prioritized because it's unclear how a severely damaging protocol change wouldn't be priced in by 28 days anyhow.
```
orangesamus, line 57: The question was here:
`proposal.vote.delay.time`<-{I'm guessing this is a setting defined somewhere else since it says "currently 1 week"... should any other context be listed for this spec?}

RPIPs/RPIP-43.md
LongForWisdom: I think your change is more correct. I'm unsure on the knock-on effects of changing just here.

RPIPs/RPIP-44.md
LongForWisdom: It was more focused on the node operator side. Eg it's probably easier for a NO to just call a smart-contract function and let RP handle it than do it themselves. 

You could add the abandonment case as well, but its explicitly mentioned in the abstract as 'not included'

RPIPs/RPIP-44.md
LongForWisdom: This is covered in the list in RPIP-43, right? You could link to RPIP-43.

RPIPs/RPIP-46.md
```diff
@@ -177,7 +177,7 @@ A few details about the reasoning behind the spec:
...
+As the core value capture is no longer based on a minimum RPL requirement, Saturn 1 removes the minimum to receive RPL issuance rewards. This "cliff" has been an extremely poor piece of UX and anecdotally led to many operators exiting validators. Note that the withdrawal limits are not updated; those had previously aligned with the end of the linear region, and they still do. With Saturn 2, RPL issuance rewards stop entirely. At this point, the bond curve should be creating sufficient attraction that we believe we can get to this desired end state more smoothly. <- [what does the "desired end state" refer to here? I'm not sure I understand what this sentence is trying to highlight"]
```
LongForWisdom, line 180: +1, last sentence doesn't quite make sense to me.

RPIPs/RPIP-47.md
```diff
@@ -31,10 +31,10 @@ This RPIP is part of a set of proposals motivated by a desire to rework Rocket P
 - Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
 - Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value
   - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days
-- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to now + its `upgrade_buffer`
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to now<-[is "now" ambiguous? I'm not sure how else you would explain this point though]  + its `upgrade_buffer`
```
LongForWisdom, line 34: 'the current block' is the probably the best alternative.

RPIPs/RPIP-59.md
```diff
@@ -77,14 +77,15 @@ A node operator MUST take 2 actions to start a validator: `deposit` and `stake`
 - `stake` SHALL revert unless at least `scrub_period` time has passed since ETH was assigned to the validator, to allow for validating the prestake
 - If the beacon chain stake is invalid, the validator SHALL be scrubbed 
 - `stake` SHALL stake the remaining 31 ETH to the beacon chain to make a complete validator
-- If `stake` is not called within `time_before_dissolve` after the ETH was assigned, the validator SHALL be dissolved, returning the unstaked balance to the deposit pool
+- If `stake` is not called within `time_before_dissolve` after the ETH was assigned [can `stake` be called by anyone? Just curious if someone could stall the entire queue for 2 weeks at a time if they wanted to (maybe they'd have no reason to and it's nbd...] , the validator SHALL be dissolved, returning the unstaked balance to the deposit pool
```
LongForWisdom, line 80: I think this might have been part of the 2tx versus 3tx discussion? Knoshua and Kane were talking about incentives to call stuff at one point, though that might have been around prestake.

RPIPs/RPIP-60.md
```diff
@@ -54,7 +54,7 @@ The explanation document is a mechanism intended to help the pDAO hold the secur
 **Upgrade Delay Setting**  
 Due to the considerations around slowing down hotfixes [described above](#security-considerations), the initial upgrade delay setting is attempting to be the smallest that allows "enough" time for participants to exit the system ahead of a proposal if they deem that appropriate.
```
LongForWisdom, line 55: Is this even still true with the setting to 1 week?

RPIPs/RPIP-43.md
Valdorff: Yeah, I think this is consistent with how we use NO elsewhere. As we use it, it's more like "node's operator". I prefer not making this change.

RPIPs/RPIP-43.md
```diff
@@ -72,14 +72,15 @@ Node operators can manage the set of validators in their megapool:
 - The oDAO SHALL be able to increase `debt` to apply penalties (see [RPIP-42](./RPIP-42.md/#penalizable-offenses)) by majority vote
   - If this causes the megapool to meet the criteria for a non-node operator initiated exit (see [RPIP-44](RPIP-44.md)), it is RECOMMENDED that an exit is initiated by the same transaction 
 - There SHALL be a function to pay off `debt` with ETH. The ETH SHALL be sent to the rETH contract and the `debt` SHALL be reduced by the amount.
+^ does the smoothing pool need to be accounted for here if they are opted into it? ^
```
Valdorff, line 75: I don't think so. debt isn't to other NOs in the smoothie, it's explicitly to rETH

RPIPs/RPIP-43.md
```diff
@@ -72,14 +72,15 @@ Node operators can manage the set of validators in their megapool:
...
 
 
 ### Funds Management
 
 In the following, "capital" refers to funds supplied for staking by either the
 node operator or pool stakers, including node operator staked RPL. "Rewards"
 refers to additional funds generated or obtained via staking, including
-consensus rewards, execution rewards, and RPL inflation. The split of rewards
+consensus rewards, execution rewards, RPL issuance, and `voter_share` rewards. The split of rewards
```
Valdorff, line 83: I don't think we need to add voter share. This is consensus and execution rewards before then being split into shares per UARS (next sentence).

Valdorff: I can take or leave issuance vs inflation

RPIPs/RPIP-44.md
Valdorff: More importantly than pure convenience, it means if you control the withdrawal address your principal is safe. Ie, if I lose my node address seed and light my NUC on fire, I am still ok.

RPIPs/RPIP-44.md
Valdorff: Agree RPIP-43 already says this and double define is bad

RPIPs/RPIP-46.md
Valdorff: the desired end state is no issuance. this is tryna say that in Saturn 1 we still need some to keep things attractive for NOs, and that in Saturn 2 that's no longer the case. too late for my brain to think of pretty words for that.

RPIPs/RPIP-47.md
```diff
@@ -1,7 +1,7 @@
 ---
 rpip: 47
 title: Enable Forced Delegate Upgrades
-description: Allows protocol governance to force upgrade node operators after a Rocket Pool protocol upgrade takes place, and a grace period has expired. 
+description: Allows protocol governance to force upgrade megapool delegates after a Rocket Pool protocol upgrade takes place, and a grace period has expired.
```
Valdorff, line 4: +1

RPIPs/RPIP-47.md
Valdorff: or even "the current block"; worth a tweak

RPIPs/RPIP-47.md
```diff
@@ -31,10 +31,10 @@ This RPIP is part of a set of proposals motivated by a desire to rework Rocket P
 - Each `Megapool delegate` contract SHALL have an execution layer `expiration_block` variable which should initialize to "no expiration"
 - Each `Megapool delegate` contract SHALL have a non-modifiable `upgrade_buffer` value
   - The `Megapool delegate`'s `upgrade_buffer` SHALL be 120 days
-- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to now + its `upgrade_buffer`
+- The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to now<-[is "now" ambiguous? I'm not sure how else you would explain this point though]  + its `upgrade_buffer`
 
 ### Upgrade Process
-- When a new `megapool delegate` is released, a protocol contract SHOULD update the `expiration_block` of all in-use `megapool delegate`s to now + their `upgrade_buffer`s
+- When a new `megapool delegate` is released, a protocol contract SHOULD update the `expiration_block` of all in-use `megapool delegate`s to now<-[same comment here about "now"?] + their `upgrade_buffer`s
```
Valdorff, line 37: +1

RPIPs/RPIP-59.md
Valdorff: This doesn't stall the entire queue, just 31 ETH worth.

Up a bit higher it looks like this is an NO-specific action "A node operator MUST take 2 actions to start a validator: `deposit` and `stake`". I think this makes sense.

RPIPs/RPIP-59.md
```diff
@@ -77,14 +77,15 @@ A node operator MUST take 2 actions to start a validator: `deposit` and `stake`
...
   - If a validator is dissolved the bonded value SHALL be recoverable. This MAY require further action from the node operator. This MAY temporarily require additional ETH from the node operator.
 
 #### Exiting Queue
 - Until ETH is assigned to a validator, it SHALL be possible to exit the queue and receive ETH `credit` for it
 
 #### Social Assignments
 - The `deposit.assign.socialised.maximum` setting SHALL be set to 0
+^ [I'm guessing this is a standard variable setting somewhere (outside of this RPIP), since I don't see it defined anywhere else here?]
```
Valdorff, line 88: this is a pre-existing variable, yes

RPIPs/RPIP-60.md
Valdorff: I think this is sufficient context. Unrelatedly, I think we should make an info RPIP with all the settings. There's a lot and some even predate RPIPs, which is fun.

RPIPs/RPIP-60.md
Valdorff: I think so -- the bit immediately following explains why I think it's ~true

RPIPs/RPIP-43.md
orangesamus: Ok, I'll update to just:
> consensus rewards, execution rewards, and RPL issuance. The split of rewards

RPIPs/RPIP-46.md
orangesamus: Maybe:
> With Saturn 2, RPL issuance rewards stop entirely. The bond curve in Saturn 2 should allow voter share to provide sufficient incentive to stake RPL, but the continuation of RPL issuance rewards under Saturn 1 should help facilitate a smoother transition to this end state

RPIPs/RPIP-47.md
orangesamus: I'll go with this unless I hear otherwise:
> The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to the current block.

RPIPs/RPIP-59.md
orangesamus: Is that 31 ETH worth per validator deposit (which is only 1.5ETH?), So if someone did: say 100 deposits of 1.5 ETH, if they wanted to they could stall 3100 ETH for two weeks? (Or even worse 31 ETH per 2 weeks, so 100*2 weeks?)

That's probably not how it works but just wanted to check

RPIPs/RPIP-47.md
Valdorff: That's wrong? Should be

> The only permitted modification to a `Megapool delegate`'s `expiration_block` variable SHALL be for any protocol contract to set its value to the current block + its `upgrade_buffer`.

RPIPs/RPIP-59.md
Valdorff: Can stall 3100 ETH for two weeks in this scenario. At the cost of making your own ETH idle and paying a buncha gas.

RPIPs/RPIP-46.md
LongForWisdom: Maybe don't call it an end state? Otherwise Samus's wording feels good.

RPIPs/RPIP-47.md
LongForWisdom: Yes, needs to be current + buffer.

RPIPs/RPIP-46.md
orangesamus: Ok, howbout:
> With Saturn 2, RPL issuance rewards stop entirely. Saturn 1 will continue to provide RPL issuance rewards to help facilitate a smoother transition, but with Saturn 2 the bond curve alone should allow voter share to provide sufficient incentive to stake RPL.

RPIPs/RPIP-47.md
orangesamus: oops, didn't realize my committed comment was mid-sentence... I'll include that in polish PR

RPIPs/RPIP-59.md
orangesamus: So the cost to stall current DP size for two weeks requires putting up (`(29,112/31)*1.5 ~= 1400 ETH`) worth of idle capital (and paying gas). I agree I don't see any incentive to do this necessarily, but should it be accounted for in security considerations? Just a (maybe outlandish) example: a competitor wants to stall rETH minting as a griefing vector, they continue to cycle 1400 ETH worth which stalls our current DP worth of rETH deposits for 2 weeks at a time each round they do it (hurts rETH APY and stalls growth)... If they had another 1400 ETH on standby they could just continue to have a constant two week stall that keeps cycling? (Maybe other deposits can get in line in between each cycle, but maybe no one wants to since NO queue is too long, especially with more ETH?) Express queue tickets could help but is limited in number.

Maybe this is already listed as a consideration:
> Disabling social assignments increases the likelihood of ETH accumulating in the deposit pool while validators are waiting in the queue. But the node operators in the queue that would be assigned ETH have a direct incentive to execute the assignments. In addition, we recommend that the pDAO funds (for example through the GMC) development and execution of assignment bots that assign ETH in the deposit pool to the queue at a reasonable gas price.

Or maybe this isn't a valid concern/too high of a cost to do this, with not enough incentive?

LongForWisdom: I don't see any harm in including it as a security consideration? I don't feel very strongly either way though.

Valdorff: I think the consideration you highlighted is different.

> If a validator is dissolved the bonded value SHALL be recoverable. This MAY require further action from the node operator. This MAY temporarily require additional ETH from the node operator.
The above is the relevant text. 

You're welcome to add a security consideration bullet. I think after one use folks would get in line and it would just be an annoyance. But nonetheless:
> There is potential for moderate griefing by creating n validators and never calling `stake`. The griefer keeps n*31 protocol ETH idle for 2 weeks at the cost of keeping n of their own ETH idle and paying gas to `deposit` and reclaim funds from `dissolve`. The text notes that reclaiming funds MAY require further action from the node operator and MAY require additional temporary ETH. Friction could be added in either of those steps should this variety of griefing prove problematic. Alternatively, a small fee to reclaim funds would also be effective (though that would require a new specification etc.).

RPIPs/RPIP-46.md
orangesamus: :rocket:

RPIPs/RPIP-47.md
orangesamus: :rocket:

RPIPs/RPIP-47.md
orangesamus: :rocket:

RPIPs/RPIP-47.md
orangesamus: :rocket:

RPIPs/RPIP-43.md
orangesamus: :rocket:

RPIPs/RPIP-59.md
orangesamus: :rocket:

PR #282, opened 2024-08-24
Tokenomics Rework Prelude

haloooloolo:
Draft RPIP and reward calculation spec for pre Saturn ETH-only minipools.


Valdorff (approved):
I gave it a once-over and am comfortable to merge as Draft.

Importantly, I did not review the reward-calculation-spec. It should be looked at before going to vote, ideally by folks that are familiar with the spec (such as @jshufro and @xrchz).

PR #283, opened 2024-09-01
RPIP-62: Change Dynamic Commission Schedule

haloooloolo:
Move from an upgrade-triggered three stage approach to two stages with interval-based transition.


Valdorff (approved):
LGTM for this stage. When we get towards finalizing, my quibble about double-specifying formulae in v10 and RPIP-62 might come up again.

PR #285, opened 2024-09-06
Fix citation link.

LongForWisdom:


Valdorff (approved):
Confirmed the capitalization fixes an example case. Ty.

PR #286, opened 2024-09-08
RPIP-62: Restrict Bonus Eligibility to LEB8s + Vote Format

haloooloolo:


Valdorff (approved):
Looks good to me.

I do still want a patches-or-ramana thumbsup before it goes to review/final. Maybe a couple others that would work too if they took the time. I'm a bit of a newbie on the actual details there.

PR #291, opened 2024-09-12
Samus RPIP63 Review

orangesamus:


RPIPs/RPIP-63.md
orangesamus, line 1: For [line 83](https://github.com/rocket-pool/RPIPs/pull/291/files#diff-abd069bc7343faf1501aa444765cfeaccf3feeb05812c26d14444dac6786d007L83) "community/oDAO verify the changes", it seems the oDAO will vote in the upgrade - for "community" verification is the idea that people will just raise a complaint socially if they find an issue? (Or will there be a snapshot vote or anything for the community to formally verify/vote approval of the changes)? I'm guessing it is just "soft/social" verification but just wanted to check...

orangesamus, line 1: I think we need to update RPIP-62 to remove the scrub penalty change since it is spec'd in RPIP-63 and will be included in the hotfix?

RPIPs/RPIP-63.md
```diff
@@ -90,4 +90,4 @@ On a failed scrub check the oDAO vote to scrub a minipool. When the scrub quorum
 
 
 ## Copyright
-Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
\ No newline at end of file
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
```
orangesamus, line 93: Also I'm not sure why github is flagging a change here :thinking:

RPIPs/RPIP-63.md
orangesamus: > I think we need to update RPIP-62 to remove the scrub penalty change since it is spec'd in RPIP-63 and will be included in the hotfix?

Done: <https://github.com/rocket-pool/RPIPs/pull/293>

darrenlangley: Yes it is a soft/social verification. It is not strictly a formal requirement but it is important so we have it in our checklist and document (via an announcement) that it was completed.

darrenlangley: > I think we need to update RPIP-62 to remove the scrub penalty change since it is spec'd in RPIP-63 and will be included in the hotfix?

Personally I would keep it in both places - imo the two RPIPs should be independent. The scrub check change is included in the hotfix for practicality but can go out without RPIP-62.

darrenlangley: > Could add the security consideration bullet to RPIP-63 that is already in RPIP-62 [here](https://github.com/rocket-pool/RPIPs/blob/main/RPIPs/RPIP-62.md?plain=1#L79C1-L79C784)

Yes good point and, in fact, I will add it to the rationale as well.

RPIPs/RPIP-63.md
darrenlangley: No idea either

kanewallmann:
Does the quorum guardrail change from 51% to 15% also need to be applied to the veto quorum or should that stay at >=51%?

kanewallmann:
Should we also include in this RPIP a change to the voting quorum setting itself? As it's currently set to 51% a proposal to lower it to 15% would require the old quorum.

RPIPs/RPIP-63.md
Valdorff: I think it's a lot simpler if it's only specified in one place (single source of truth and have a requirement in the other). If scrub stuff isn't in the hotfix, I think RPIP-62's goals would probably be attempted pretty differently.

Valdorff:
> Does the quorum guardrail change from 51% to 15% also need to be applied to the veto quorum or should that stay at >=51%?

I could argue it makes more sense lower (say 25% or 33%), but it's not breaking anything at 51% so I probably wouldn't address it in the hotfix tbh. Only relevant if we see cases of spam that aren't able to hit veto quorum.

> Should we also include in this RPIP a change to the voting quorum setting itself? As it's currently set to 51% a proposal to lower it to 15% would require the old quorum.

Right -- we'd need a single vote at the old quorum to update. I think we're at about 50k/94k, so we could only lower it as far as 28.2% cleanly. I do think it's worth including a reduction to 30% or something. That would still be over the desired 15% end state, and also makes it much easier to act in the meanwhile.

Valdorff:
Note the merged https://github.com/rocket-pool/RPIPs/pull/295/ adds in quorum change to 30%

orangesamus:
Closing since the conversations were accounted for in other PR's:
<https://github.com/rocket-pool/RPIPs/pull/293>
<https://github.com/rocket-pool/RPIPs/pull/294>
<https://github.com/rocket-pool/RPIPs/pull/295>

PR #297, opened 2024-09-21
RPIP-63 - Added knoshua as author

darrenlangley:


Valdorff:
The desired change was already included in https://github.com/rocket-pool/RPIPs/pull/296 so this PR actually includes no changes. Closing.

PR #301, opened 2024-09-30
RPIP-63: setting to final ready for vote

darrenlangley:


xrchz (approved):
Although I don't think it's a good idea to bundle these changes, I don't think it's an RPIP editor job to make that call at the moment.

PR #303, opened 2024-10-03
Update RPIP-29 for October 2024

shfryn:
I haven't updated this in a while due to updates in the new treasury sheet, which have made the older sheet difficult to maintain. The RPIP-29 format now requires an overhaul, as it's no longer maintainable without a significant time investment and it’s much more difficult for users to read and understand. Originally designed to handle 5-10 transfers per month, the GMC has recently been averaging around 50 transfers per month. We need a more efficient way to track ongoing projects to make them easier to manage, but the ideal format for this still needs to be determined.


Valdorff (approved):

Valdorff:
Focused verification on (a) matching GMC Treasury Sheet and (b) open grants/bounties

PR #306, opened 2024-10-10
Update 005-glossary.md with new terms

paladin147:
Added smoothie and rewards tree


Valdorff:
@paladin147 please include the fact that smoothing pool payouts are calculated offchain and use the rewards tree. Could be under either or both definitions.

PR #310, opened 2024-10-31
Updated for October 2024

shfryn:


Valdorff (approved):
All looked reasonable with one exception: the weekly orbit amount to be paid didn't change but the last paid date did. Shfryn confirmed that there was a miscalculation in the previous iteration and that it was correct as shown this time.

PR #315, opened 2025-01-02
december 2024 update

shfryn:


xrchz (approved):
Haven't checked all the details, but approve of these updates in general.

PR #336, opened 2025-04-29
Finalize RPIP-70

Valdorff:
- Grammarly
- discussions-to link set to N/A in header; discord and forum had bits of it, but there was no good single resource
- Fix description in header
- Set to final


htimsk (approved):
Simple updates.

PR #337, opened 2025-05-14
Add draft of withdrawal liquidity RPIP

xrchz:


Valdorff (approved):
- [x] Ensure that the RPIP preamble is present and correctly formatted as per RPIP-1.
  - [x] Add RPIP number if not present.
  - [x] Check `title` is a few words, and not a complete sentence.
  - [x] Check `author` field matches convention.
  - :x: Check `discussions-to` field is present, and links to the Rocket Pool forum. This field should consist of a single URL.
  - [x] Check `status` is set to 'Draft'
  - [x] Check that `type` is a valid value.
    - [x] If the `type` is 'protocol', ensure a `category` is present. 
  - [x] Check that `created` is set to the day of the RPIP number assignment, in the YYYY-MM-DD format.
- [x] Ensure that the RPIP matches the RPIP template in terms of layout.
- [x] Ensure that the RPIP author is content with the draft being published on the RPIPs portal in its current state.

Looks good enough -- can fill in the missing bit as this matures.

PR #348, opened 2025-07-31
Rpip73 reth protection

orangesamus:


MR-Neto:
Great work! 🎖️ 

I had thought in the past of some insurance from staked RPL to rETH yield. It would be something like reduce RPL eth commission if rETH return is below some threshold of an expected yield.

htimsk (approved):
Creating RPIP for rETH Protection

PR #370, opened 2026-05-08
First draft

knoshua:


RPIPs/RPIP-80.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Specification
+
+This specification introduces the following pDAO protocol parameters:
+
+| Name                     | Type  | Initial Value | Guardrail<br> |
+| ------------------------ | ----- | ------------- | ------------- |
+| `cooperative_exit_phase` | Hours | 72            |               |
+| `no_exit_penalty`        | ETH   | 0.1           |               |
+| `no_exit_cooldown`       | Days  | 28            | > 7           |
```
Valdorff: If I'm reading this right, both of these variables apply per-validator. I suggest renaming them to `validator*` instead of `no*`

RPIPs/RPIP-80.md
```diff
@@ -0,0 +1,89 @@
...
+
+- The protocol SHALL expose a variable `requested_eth` that tracks ETH expected to exit under this specification. 
+- For the purpose of this specification expected user capital is defined as `32 ETH - validator bond`.
+
+### Exit Requests
+
+- The protocol SHALL be able to request a validator to exit for underperformance (RPIP-73) or withdrawal liquidity (RPIP-71). The protocol MAY provide an interface for future protocol contracts to request exits.
+- When a validator is requested to exit, the protocol MUST expose an “exit requested” signal that is observable to the node operator and `requested_eth` SHALL be increased by the expected user capital of that validator.
+- After `no_exit_cooldown` has passed since a request, the validator SHALL no longer be considered requested to exit and can be requested to exit again.
+- Requesting a validator that is already considered requested to exit SHALL not reset the window or add to `requested_eth`.
```
Valdorff: minor: `SHALL NOT` with the not capitalized

RPIPs/RPIP-80.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Delegates Incorporating EIP-7002
+
+- A new minipool delegate SHALL expose an interface that allows initiating an execution-layer-triggered exit under the conditions defined in this specification.
+- The exit functionality of megapools introduced in RPIP-44 SHALL be extended to allow initiating an execution-layer-triggered exit under the conditions defined in this specification.
+
+### Tracking Requested ETH
+
+- The protocol SHALL expose a variable `requested_eth` that tracks ETH expected to exit under this specification. 
+- For the purpose of this specification expected user capital is defined as `32 ETH - validator bond`.
```
Valdorff, line 43: Good spec work. I was wondering if we'd run into trouble if there was leakage or somesuch, but this concrete definition is accurate enough and doesn't suffer from such issues.

RPIPs/RPIP-80.md
```diff
@@ -0,0 +1,89 @@
...
+That's why the proposal uses the third option. A penalty for failing to exit is meant to provide an economic incentive for protocol-aligned behavior: upgrading to the new delegate or responding to exit requests.
+
+### Limiting Minipool Penalty Frequency 
+
+A validator that was requested to exit  in the last `no_exit_cooldown` days cannot be requested to exit again. This is relevant for the case of minipools that can't be exited and limits the frequency with which penalties can be applied, giving the node operator an opportunity to upgrade instead of quickly ramping up the penalty.
+
+### Penalty Sizing
+
+The specification uses an absolute penalty value, meaning that 16 ETH and 8 ETH minipools receive the same penalty per failed exit. Arguments could be made for scaling either way. If we wanted to ensure that both cases experience the same loss of APR, the penalty for 16 ETH should be scaled up. From the protocol's point of view, the 8 ETH minipool failing to exit 24 ETH of user capital is more damaging than the 16 ETH case and arguably should be penalized more.
```
Valdorff: > From the NO's point of view, the flat penalty is more damaging to their APR for the 8 ETH minipool than the 16 ETH minipool.

Your later sentence is clearer, so I'd suggest changing this one to match

Valdorff: I'd probably also suggest one more resolving sentence like "As such, we've chosen to take the compromise position of an absolute penalty, which is neither fully right nor maximally wrong from either point of view."

RPIPs/RPIP-80.md
knoshua: "no exit" wasn't supposed to  mean node operator here, but "opposite of did exit". I also confused myself while going over my draft, so probably a good indication that it's not great naming in RP context. Changed it to something hopefully more clear.

Valdorff: lol - fairnuff and better now

Valdorff (approved):
LGTM

PR #371, opened 2026-05-08
rETH Protection From Underperforming Nodes - Updated Draft

knoshua:


Valdorff (approved):
This is great.

RPIPs/RPIP-73.md
```diff
@@ -2,55 +2,74 @@
...
+| `performance_period`           | Epochs | 22016 (~100 days) |               |
+| `performance_threshold`        | pct    | 94                |               |
+| `performance_challenge_period` | Hours  | 24                |               |
+| `performance_challenge_bond`   | RPL    | 100               | > 20          |
+
+A * designates this parameter as being modifiable by the Security Council without a delay.
+
+### Performance Challenge Mechanism
+
+- If `performance_exits_enabled` is `true` , the protocol SHALL allow anyone to propose a validator exit by providing a `start_epoch` and `performance_period * (1 - performance_treshold)`epochs within `[start_epoch, start_epoch + performance_period]` and locking `performance_challenge_bond` for `performance_challenge_period`.
```
Valdorff, line 41: > `performance_period * (1 - performance_treshold)`epochs

This is probably sufficient for specification, but in a rationale or discussion section I'd probably address how to actually handle the input, as I believe the naive approach is an array well over 1000 elements long. My offhand thought is to allow that and also allow an array where odd numbers are epochs and even numbers are number of contiguous epochs -- I suspect very many of these cases will be large blocks of bad time and we might as well save some gas.

knoshua: Kane actually had some good ideas on this already. I was trying to keep it vague and high-level in the specification to keep the RPIP readable for a wide audience and to not restrict what the team can do when implementing, in case they come up with something even better.

But I like your idea of addressing it a bit outside of the Specification section and added a paragraph. I think you are suggesting something along the lines of submitting interval borders of periods of not-timely votes, but wasn't completely sure I got it. Feel free to add your own bullet to the section before merging. Also open to edits to the spec bullet to better indicate that details are meant as an implementation detail.

PR #381, opened 2026-06-08
Update RPIP-79

knoshua:


RPIPs/RPIP-79.md
```diff
@@ -29,28 +30,26 @@ This specification changes the following pDAO protocol parameter guardrails:
...
 
-## References
-- [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) introduced precompiles for BLS12-381 curve operations that allow signature verification.
-
 ## Security Considerations
-A fraud proof only succeeds if another deposit with the same pubkey and valid signature exists. Compared to the legacy oDAO scrub system and the beacon state proof system, the threat model is therefore essentially unchanged: an adversary who cannot sign as the validator cannot cause dissolution. The main caveat introduced by this proposal is that back-running with incorrect withdrawal credentials also leads to dissolution.
+A fraud proof only succeeds if another deposit with the same pubkey and a valid signature exists. Compared to the legacy oDAO scrub system and the beacon state proof system, the threat model is therefore essentially unchanged: an adversary who cannot sign as the validator cannot cause dissolution. The main caveat introduced by this proposal is that back-running with incorrect withdrawal credentials also leads to dissolution.
```
Valdorff: > There is one new path to dissolution, which is a valid signature and incorrect withdrawal credentials that happens after a correct first deposit; as mentioned previously, there is no legitimate reason for such a deposit to exist and dissolving is the conservative choice for the protocol.

This was not clear to me, but your earlier wording was. Maybe could reword to build off of that? Eg:

> There is one new path to dissolution, which is a valid signature and incorrect withdrawal credentials that happens after a correct first deposit; as mentioned previously, there is no legitimate reason for such a deposit to exist and dissolving is the conservative choice for the protocol.

knoshua: Updated that section to hopefully make it clearer

Valdorff (approved):

PR #382, opened 2026-06-10
RPIP-71

knoshua:


RPIPs/RPIP-71.md
```diff
@@ -2,33 +2,209 @@
...
+- Canceling a withdrawal request and exiting the queue SHALL NOT be possible.
+
+### Fulfilling Withdrawal Requests
+
+- While the Withdrawal Queue is not empty:
+	- ETH from new rETH mints SHALL first go towards the Withdrawal Queue.
+	- Redemption of rETH outside the queue SHALL be prevented to the extent possible.
+	- The protocol SHALL allow the Withdrawal Queue to burn rETH.
+- `network.reth.collateral.target` SHALL be set to 0 and a new buffer for withdrawals SHALL be implemented in the deposit pool, reserving up to `deposit_pool_collateral_target` percent of ETH backing rETH for rETH burns.
+- When megapools distribute rewards, they SHALL send ETH to the deposit pool rather than rETH.
```
Valdorff: Hmmm... clear to me, but maybe more explicit like: `rather than the rETH token address` or `rather than the rETH contract`

RPIPs/RPIP-71.md
```diff
@@ -2,33 +2,209 @@
...
+- `requested_exit_eth` (defined in RPIP‑80),
+- `voluntary_exit_eth` (defined in RPIP-80),
+- ETH held by the rETH contract, and
+- ETH held by the deposit pool.
+
+If the withdrawal shortfall is at least `exit_hysteresis` ETH, additional megapool validators MUST be selected and added to the exit list as defined in this section.
+
+#### If `megapool_exit_phase = false`
+
+- The protocol SHALL allow the oDAO, by majority vote, to request minipools to exit as specified in RPIP-80. The amount necessary to cover the withdrawal shortfall minus  `exit_hysteresis` SHALL act as a hard cap on the possible exit requests.
```
Valdorff: > minus  `exit_hysteresis` 
typo double space

RPIPs/RPIP-71.md
```diff
@@ -2,33 +2,209 @@
...
+
+### Validator Exit Selection
+
+For the purposes of this section, the withdrawal shortfall is defined as the remaining ETH required to fully satisfy all pending Withdrawal Queue requests after accounting for:
+- `requested_exit_eth` (defined in RPIP‑80),
+- `voluntary_exit_eth` (defined in RPIP-80),
+- ETH held by the rETH contract, and
+- ETH held by the deposit pool.
+
+If the withdrawal shortfall is at least `exit_hysteresis` ETH, additional megapool validators MUST be selected and added to the exit list as defined in this section.
```
Valdorff, line 68: For author consideration: Would another way to handle this be based on request age? I think that would allow more intuitive statements (eg, enough ETH to service your request will be exited after x days), and it would also be an easier pdao knob to set (eg, we are comfortable with folks waiting y days [beyond queues] in order to avoid churn). 

I have not been following closely, so this may have been considered and discarded.

RPIPs/RPIP-71.md
```diff
@@ -2,33 +2,209 @@
...
+- When `notifyExit` is called for a megapool validator that is still in the eligible megapool validator set, that validator MUST be removed or the scheduled inclusion MUST be canceled.
+
+##### Tournament-based selection
+
+Let `N` be the current size of the eligible megapool validator set. Let `k` be `tournament_size` (or `N` if `tournament_size > N`). The protocol SHALL allow repeated selection of validators while:
+- the withdrawal shortfall is at least `exit_hysteresis` ETH, and
+- the eligible megapool validator set is non‑empty.
+
+1. The protocol MUST sample `k` distinct validators uniformly at random without replacement from the eligible megapool validator set.
+2. For each sampled validator, the protocol MUST compute an exit score according to the chosen Exit Criterion.
```
Valdorff, line 96: Is Exit Criterion defined somewhere? It should either be defined, or responsibility for defining should be given to an entity

RPIPs/RPIP-71.md
```diff
@@ -2,33 +2,209 @@
...
+
+Using the oDAO for minipools also allows us to order exits by commission, which marginally improves the rETH APR in response to rETH demand reduction.
+
+We focus on minipools in the first phase, not only to keep the reliance on the oDAO as short as possible, but also because this aligns with the overall desire to move away from minipools and towards megapools: to support the new RPL tokenomics, the new Universal Adjustable Revenue Split mechanism ([RPIP-46](RPIP-46.md)), and to reduce complexity of future upgrades.
+
+We also considered a trustless design for minipools. This would increase complexity and incur additional gas costs that the protocol or pDAO would either have to fund in some way or add to oDAO's duties. The trusted oDAO duty appears preferable because of its temporary nature, limited risk, reduced complexity and lower gas cost.
+
+The `megapool_exit_phase` toggle allows the pDAO to switch to the second phase once minipools are largely exited or known to be unresponsive. This choice is left to the pDAO because we cannot precisely define it in advance.
```
Valdorff: If they are known unresponsive, shouldn't they be getting tried on cooldown and penalized if appropriate? I'm wondering if the megapool exit phase should include requesting exits on any remaining minipool every `did_not_exit_cooldown` or something.

RPIPs/RPIP-71.md
Valdorff: I think it's also a bit stronger against the "Repeated Minting and Withdrawing" scenario, though I don't really think further strength there is critical

RPIPs/RPIP-71.md
knoshua: I was thinking that the pDAO at that point could come up with a plan for any remaining minipools. I'm not sure it would make sense to keep going with small penalties over time, but we could vote to have oDAO apply one (bigger) penalty,  leaving some bond left over so that those NOs still have an incentive to exit. We could definitely come up with a plan now as well, I just figured this isn't a big priority and we may have a better idea in the future.

RPIPs/RPIP-71.md
knoshua: Good point. Earlier versions had an "open questions" section that made it clear that exit criterion still needs to be figured out. That got dropped and now it's not clear at all where exit criterion is going to come from. Adding a section to Rationale.

My expectation right now is that we'll have pDAO vote on it and update the specification section based on outcome of the vote. Maybe the RPIP should reflect that directly in the specification?

RPIPs/RPIP-71.md
knoshua: I'll have to think on that.

Valdorff (approved):
LGTM - my concerns were addresses, as was a discussion on correlated slashings

PR #383, opened 2026-06-26
RPIP-80 Update

knoshua:


Valdorff (approved):
Scaling time and penalty together sensibly 👍

PR #384, opened 2026-07-08
RPIP-73 Update

knoshua:
- allow challenging a list of validators
- ensure that only epochs where the validator should have been attesting can be challenged
- add a buffer to give time to detect under-performance and generate proofs


Valdorff (approved):
Lists supported and clarity improved 👍

PR #385, opened 2026-07-08
RPIP: On chain signal voting

knoshua:


Valdorff (approved):
LGTM. Simple minimal design for the onchain component, assuming a frontend exists to help interact.

PR #390, opened 2026-07-21
RPIP for new signal voting platform

DrDoofus-MD-PhD-DDS:
This RPIP creates a new Recognized Signaling Platform and implements RocketDash/vote as the initial RSP.
It puts langers as the initial Lead Vote Administrator.
Finally, it creates an Emergency Ballot System to help us move RSPs if the current one becomes unusable or untrusted.


Valdorff (approved):
Looks fine on quick skim. Should get more thorough review before Final though.

PR #392, opened 2026-07-28
RPIP-36: Update RPIP Editors

knoshua:


Valdorff (approved):
Update is accurate plus or minus formal acceptance making the date a bit ambiguous.

PR #395, opened 2026-07-28
RPIP-79: Add additional proof to cover invalid signature case

knoshua:


Valdorff (approved):
straightforward technical improvement

